PR #167, opened 2015-01-11
Feature/translators

friedger:
Adding a list of translators and a section about contributing to localizations

**After**
screenshot_2015-01-11-20-43-32
screenshot_2015-01-11-20-43-28

Note, the actionbar is not fixed in the PR.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
@@ -16,100 +16,22 @@
 
 package org.gdg.frisbee.android.fragment;
 
-import android.content.Intent;
-import android.net.Uri;
-import android.os.Bundle;
-import android.view.LayoutInflater;
-import android.view.View;
-import android.view.ViewGroup;
-import android.widget.ListView;
-
-import com.android.volley.Response;
-import com.android.volley.VolleyError;
-
-import org.gdg.frisbee.android.Const;
-import org.gdg.frisbee.android.R;
-import org.gdg.frisbee.android.adapter.ContributorAdapter;
-import org.gdg.frisbee.android.api.ApiRequest;
-import org.gdg.frisbee.android.api.GitHub;
 import org.gdg.frisbee.android.api.model.Contributor;
-import org.gdg.frisbee.android.app.App;
-import org.gdg.frisbee.android.cache.ModelCache;
 import org.gdg.frisbee.android.crowdin.model.Translator;
-import org.gdg.frisbee.android.utils.Utils;
-import org.joda.time.DateTime;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-
-import butterknife.ButterKnife;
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.gdg.frisbee.android.api.model.Contributor;
+import org.gdg.frisbee.android.crowdin.model.Translator;
+
+
+public class TranslatorsFragment extends ContributorsFragment {
+
+    private static final Contributor[] TRANSLATORS = new Contributor[]{
+            new Translator("friedger", "https://www.gravatar.com/avatar/72c15c247727ba65f72de3c7c58c4a42?s=170&d=https%3A%2F%2Fcrowdin.com%2Fimages%2Fuser-picture.png", 103),
+            new Translator("tasomaniac", "https://www.gravatar.com/avatar/67be1b058c66ed002ff45a1f9a22c0ff?s=150&d=https%3A%2F%2Fcrowdin.com%2Fimages%2Fuser-picture.png", 35),
```
tasomaniac: So, we will manually update this on every update. What maybe the other options?

app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
```diff
@@ -0,0 +1,37 @@
...
+            new Translator("mauimauer", "https://www.gravatar.com/avatar/b9236795d95774ca2137bb15d54da0a9?s=150&d=https%3A%2F%2Fcrowdin.com%2Fimages%2Fuser-picture.png", 13),
+            new Translator("Andrulko", "", 2)
+    };
+
+    protected void loadContributors() {
+        for (Contributor contributor : TRANSLATORS) {
+            mAdapter.add(contributor);
```
tasomaniac: `Collections.addAll()` maybe?

tasomaniac:
:+1:

app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
friedger: `Collections.addAll()` does not work with `Adapter` and `ArrayAdapter.addAll()` is only available in [API level 11](http://developer.android.com/reference/android/widget/ArrayAdapter.html#addAll(java.util.Collection<? extends T>)). So we would need to increase the current minSDKLevel 10 - probably not now.

app/src/main/java/org/gdg/frisbee/android/fragment/TranslatorsFragment.java
friedger: The is a pending [feature request at crowdin](http://crowdin.uservoice.com/forums/31787-collaborative-translation-tool/suggestions/4734164-add-an-api-to-retrieve-a-list-of-translators).

Currently, I don't see any other option.

PR #169, opened 2015-02-02
Fix broken build because of missing cardview-v7 lib.

HenriqueRocha:
Add cardview-v7 to build.gradle to fix "No resource identifier found for... attribute 'cardCornerRadius'" compilation error.


friedger:
:+1: Thanks!

PR #175, opened 2015-02-06
Granular play services libraries are used.

tasomaniac:
It builds and be able to run on my local.

I see the welcome screen but it couldn't get chapters list. I think it is because I dont have keys. 

Lets see if it builds on Travis.


tasomaniac:
Thank you

tasomaniac:
Why is it still failing on Travis

HenriqueRocha:
I also can't get the chapter list in the production version that I just installed from the Play Store.

PR #179, opened 2015-02-07
Migrate to android environment on Travis rather than java

stefanhoth:
This PR moves the travis build from Java environment to the android one, currently still in beta.

This will hopefully will reduce the build time significantly as there is almost no download required.

More info can be found here: http://docs.travis-ci.com/user/languages/android/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ android:
     # - sys-img-x86-android-21
 
 before_script:
-    - GRADLE_OPTS=-Xmx512m -XX:MaxPermSize=256m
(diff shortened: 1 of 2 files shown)


stefanhoth:
After raising the memory to the maximum of 1.5GB I could finally get the build to pass.

If we can start using this config we can rebase the open PRs onto this and make the builds pass.

cc @friedger @mauimauer @tasomaniac

tasomaniac:
I love the gif :) 

I saw the green checkmark and the changed lines really makes sense. I can rebase my PR after this merge

PR #181, opened 2015-02-08
Trying to speed up the build on travis by caching remote dependencies

stefanhoth:
:x:  Please don`t this yet. Trying out if the new config with caching dependencies between builds brings a speed improvement.


stefanhoth:
The config didn't seem to speed up the build as the gradle wrapper and the gradle deps are still being downloaded.

Issue #182, opened 2015-02-08
Remove license text from individual files


friedger:
Remove all license text from individual files that has been created by this project.

HenriqueRocha:
Instead of removing, shouldn't we just create one single commit that updates the copyright of all the files?
It's a pretty straightforward "replace all".

tasomaniac:
Should we close this?

HenriqueRocha:
@friedger This can also be closed since you've already merged #187 :+1:

PR #187, opened 2015-02-09
Update Copyright to 2015

HenriqueRocha:
Instead of removing the copyright from all the files as suggested in issue #182 I've done a "replace all" to change the copyright from 2013 to 2013-2015. Keeping the copyright in all the files doesn't seem that bad and it only requires one "replace all" and one commit per year to update.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/Const.java
+++ b/app/src/main/java/org/gdg/frisbee/android/Const.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 The GDG Frisbee Project
+ * Copyright 2013-2015 The GDG Frisbee Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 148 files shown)


ViliusKraujutis:
:+1:

ViliusKraujutis:
NB: includes couple import lines rearranges.

HenriqueRocha:
Yes, I've noticed that also. Strange...

ViliusKraujutis:
That's no problem. That's probably your IDE settings are to rearrange/optimize imports.

HenriqueRocha:
I had those checkboxes unchecked. Maybe it was because I was working on the new pull to refresh and realized "Oh, I should update the copyright" and forgot to stash those changes.

stefanhoth:
@HenriqueRocha As @friedger already mentioned the copyright and license info should be removed from all files rather than updated. We have the [LICENSE FILE](https://github.com/gdg-x/frisbee/blob/develop/LICENSE) to cover the whole project.

Or do I miss anything here?

HenriqueRocha:
@stefanhoth As I commented in the first comment here, I proposed updating the copyright date of all files since it's one commit per year and then we keep the copyright in all the files.

The Apache Licence http://www.apache.org/dev/apply-license.html states the following:
Each original source document (code and documentation, but excluding the LICENSE and NOTICE files) SHOULD include a short license header at the top.

Since it doesn't say MUST, if you and @friedger consider that this PR is not relevant you can close it.

friedger:
I am not an expert on licences, as the APL2 is very open anyway I don't think it matters a lot. I am happy with both solutions.

Issue #190, opened 2015-02-10
Choose between Volley or Picasso for image loading


HenriqueRocha:
Right now the app uses both Volley and Picasso for image loading.
Which one do you think we should use and why?
I'll remove the one we exclude.

friedger:
Volley is used for the API calls, hence, I suggest to remove Picasso.


  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-02-11 0:08 GMT+01:00 Henrique Rocha <notifications@github.com>:

tasomaniac:
I don't like how Volley handles image loading personally. There are even
libraries that make it look like Picasso api. Picasso is really lightweight
to use.

friedger:
What do you don't like in particular? We can write a wrapper if needed or
so. We only need default behavior...
Am 11.02.2015 10:00 schrieb "Said Tahsin Dane" <notifications@github.com>:

HenriqueRocha:
I also don't like Volley's API for dealing with images but I believe that removing one of them will probably end the out of memory errors that I sometimes get. I suppose it's because the app is using two different bitmap caches.

mauimauer:
I would encourage a move to OkHttp + Retrofit + Picasso, for Image Loading and all other API implementations.

HenriqueRocha:
:+1: to use Square libs

HenriqueRocha:
Pull request #192 should close this issue by moving to Picasso.

PR #192, opened 2015-02-11
Remove Chris Bane's Bitmap Cache library

HenriqueRocha:
To address issue #190 I've removed Chris Bane's Bitmap Cache library from the project and now only Picasso is used for loading images.
When I have more time I'll also change from Volley to Retrofit but I'll work on other things first.

### DIFF
--- a/app/src/main/assets/third_party.html
+++ b/app/src/main/assets/third_party.html
@@ -140,21 +140,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.<br/>
 See the License for the specific language governing permissions and<br/>
 limitations under the License.<br/><br/>
 
-<b>Android-BitmapCache</b><br/>
-Copyright 2011, 2012, 2013 Chris Banes<br/><br/>
-
-Licensed under the Apache License, Version 2.0 (the "License");<br/>
-you may not use this file except in compliance with the License.<br/>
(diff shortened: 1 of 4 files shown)


app/src/main/assets/third_party.html
```diff
@@ -140,21 +140,6 @@
...
-You may obtain a copy of the License at<br/><br/>
-
-http://www.apache.org/licenses/LICENSE-2.0<br/><br/>
-
-Unless required by applicable law or agreed to in writing, software<br/>
-distributed under the License is distributed on an "AS IS" BASIS,<br/>
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.<br/>
-See the License for the specific language governing permissions and<br/>
-limitations under the License.<br/><br/>
-
```
stefanhoth: :+1: that you thought of that

stefanhoth:
Good effort but I'm a bit puzzled because you didn't change any code that used Volley for image loading. Was that part already replaced before?

HenriqueRocha:
@stefanhoth Yes, I was a bit puzzled myself when I removed the code from GdgVolley class.
It was initializing the bitmap cache but not actually using it for anything at all.

stefanhoth:
Weird. Thanks for removing dead code then. :+1:

PR #194, opened 2015-02-13
Fix bug that crashed the app when moving to the second step of the first start flow

HenriqueRocha:
* Remove all Robotium tests which were not being used
* Add Espresso library
* Add test that checks if the Confirm button moves to the 2nd screen of the first start flow
* Fix initial flow bug to make the test pass

The bug was caused by a previous refactoring that removed the newInstance method from the FirstStartStep1Fragment that was adding an empty Bundle to the arguments, which means that the code itself needs more refactoring but we should fix the bug ASAP.


tasomaniac:
Should we merge this. I think it looks ok and fixes the crash. 

Actually, the following line that causes the crash can be removed as well. It doesn't actually do anything. 

    getArguments().putParcelable("selected_chapter", selectedChapter);

HenriqueRocha:
I've just pushed a commit that takes advantage of what @tasomaniac wrote to better fix the crash.

tasomaniac:
I think we should merge this to continue other things. This crash is kinda a blocker for other issues.

HenriqueRocha:
@tasomaniac I also think this should be merged ASAP.
@stefanhoth or @friedger should look into this PR first when they have time.

app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
```diff
@@ -326,7 +326,9 @@ public void onErrorResponse(VolleyError volleyError) {
         public FirstStartPageAdapter(FragmentManager fm) {
             super(fm);
             mFragments = new Fragment[]{
-                    new FirstStartStep1Fragment(), new FirstStartStep2Fragment(), new FirstStartStep3Fragment()
+                    new FirstStartStep1Fragment(),
+                    new FirstStartStep2Fragment(),
+                    new FirstStartStep3Fragment()
```
stefanhoth: :+1:

stefanhoth:
LGTM, :ship: :it:

app/src/main/java/org/gdg/frisbee/android/fragment/FirstStartStep1Fragment.java
```diff
@@ -138,7 +138,6 @@ public void onNotFound(String key) {
             @Override
             public void onClick(View view) {
                 Chapter selectedChapter = (Chapter) mChapterSpinner.getSelectedItem();
-                getArguments().putParcelable("selected_chapter", selectedChapter);
```
friedger: What happens when the device is rotated?

tasomaniac: Arguments are never used elsewhere in the Fragment. Selected item in the Spinner will probably reset when device is rotated. We can make a separate issue for that. 
And for state restoration, savedInstanceState should be used. Using Arguments for that is wrong.

PR #195, opened 2015-02-16
Gradle dependencies tidied.

tasomaniac:
Butterknife and Otto are updated

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -93,7 +93,6 @@ dependencies {
     //Google
     compile 'com.google.code.gson:gson:2.3'
     compile 'com.google:volley:4.3'
-    compile 'com.google.android.analytics:analytics:3'
     compile 'com.google.android:youtube-android-player-api:1.0.0'
     compile 'com.google.http-client:google-http-client-gson:1.18.0-rc'
     compile 'com.google.apis:google-api-services-plus:v1-rev126-1.18.0-rc'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,42 +77,51 @@ android {
 }
 
 dependencies {
+    //Android Support
     compile 'com.android.support:support-v4:21.0.3'
-    compile 'com.viewpagerindicator:library:2.4.1@aar'
-    compile 'com.google.code.gson:gson:2.3'
-    compile 'org.jsoup:jsoup:1.7.3'
-    compile 'joda-time:joda-time:2.3'
-    compile 'com.jakewharton.timber:timber:2.3.0'
-    compile ('de.keyboardsurfer.android.widget:crouton:1.8.4') {
-        exclude module: 'support-v4'
-    }
-    compile 'com.google:volley:4.3'
-    compile 'com.google.android.analytics:analytics:3'
-    compile 'com.jakewharton:disklrucache:2.0.2'
-    compile 'com.google.android:youtube-android-player-api:1.0.0'
+    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.android.support:cardview-v7:21.0.3'
+    compile 'com.android.support:mediarouter-v7:21.0.3'
 
+    //Play Services
     compile 'com.google.android.gms:play-services-location:6.5.87'
     compile 'com.google.android.gms:play-services-games:6.5.87'
     compile 'com.google.android.gms:play-services-identity:6.5.87'
     compile 'com.google.android.gms:play-services-plus:6.5.87'
     compile 'com.google.android.gms:play-services-appstate:6.5.87'
-    
+
+    //Google
+    compile 'com.google.code.gson:gson:2.3'
+    compile 'com.google:volley:4.3'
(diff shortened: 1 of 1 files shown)


HenriqueRocha:
:+1:

tasomaniac:
I was also trying to remove unnecessary things. I have just pushed another commit which removes analytics dependency. It is already included in play services. It compiles and it should send data to analytics. 
It will be better if someone with access can check. Thank you.

app/build.gradle
```diff
@@ -71,47 +71,55 @@ android {
...
+    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.android.support:cardview-v7:21.0.3'
+    compile 'com.android.support:mediarouter-v7:21.0.3'
 
+    //Play Services
     compile 'com.google.android.gms:play-services-location:6.5.87'
     compile 'com.google.android.gms:play-services-games:6.5.87'
     compile 'com.google.android.gms:play-services-identity:6.5.87'
     compile 'com.google.android.gms:play-services-plus:6.5.87'
     compile 'com.google.android.gms:play-services-appstate:6.5.87'
```
stefanhoth: Random comment: Would be nice to use an ext. variable here for the GMS version.

app/build.gradle
```diff
@@ -71,47 +71,55 @@ android {
...
+    compile 'com.squareup:otto:1.3.6'
     compile 'com.squareup.okhttp:okhttp:1.6.0'
     compile 'com.squareup.okhttp:okhttp-urlconnection:1.6.0'
-    compile 'com.nineoldandroids:library:2.4.0'
     compile 'com.squareup.picasso:picasso:2.3.0'
-    compile 'com.jakewharton:butterknife:5.1.0'
-    compile 'com.google.zxing:android-integration:3.1.0'
-    compile 'com.google.zxing:core:3.1.0'
+
+    //Jake Wharton
```
stefanhoth: Lol, personal section. :+1:

app/build.gradle
```diff
@@ -71,47 +71,55 @@ android {
...
-    compile 'com.android.support:cardview-v7:21.0.3'
-    compile 'com.android.support:mediarouter-v7:21.0.3'
     compile 'com.google.android.apps.dashclock:dashclock-api:2.0.0'
-    compile 'com.squareup:otto:1.3.4'
+    compile 'com.google.zxing:android-integration:3.1.0'
+    compile 'com.google.zxing:core:3.1.0'
+
+    //Square
+    compile 'com.squareup:otto:1.3.6'
     compile 'com.squareup.okhttp:okhttp:1.6.0'
```
stefanhoth: 2.x has already been released. Major changes in the API?

app/build.gradle
tasomaniac: Is there an easy way to do it?

app/build.gradle
tasomaniac: I think it introduces major changes. Before 2.x, it was used just like HTTPUrlConnection but there were magic underneath. But after 2.x, API changed.

app/build.gradle
tasomaniac: He deserves it :smile:

app/build.gradle
stefanhoth: Ok, let's leave this for now.

app/build.gradle
stefanhoth: Check this http://tools.android.com/tech-docs/new-build-system/tips#TOC-Controlling-Android-properties-of-all-your-modules-from-the-main-project.

stefanhoth:
During testing with deleted preferences I found that I couldn't sign-in with G+ anymore during onboarding. Can someone verify?

tasomaniac:
What is the seen problem? I cannot verify since I don't have the keys.

HenriqueRocha:
I tested OkHttp 2.2.0 and everything works.
I created a PR #204 with the updates.

HenriqueRocha:
@stefanhoth It's also not working for me when I test on the phone instead of the emulator. I'll investigate the problem, hopefully fix it and submit a PR.

HenriqueRocha:
@stefanhoth I think bug was introduced in commit f1ff725 by @mauimauer a long time ago (which is kind of strange). The app is now crashing because it is doing a `getString()` in the `SharedPreferences` and the value that is put there is of type `Chapter`.

> Key org.gdg.frisbee.CHAPTER expected String but value was a org.gdg.frisbee.android.api.model.Chapter.  The default value <null> was returned.
> Attempt to cast generated internal exception:
> java.lang.ClassCastException: org.gdg.frisbee.android.api.model.Chapter cannot be cast to java.lang.String

The app only crashes if you sign in. I will work on a fix for this during this week and create a PR with all the details.

HenriqueRocha:
@stefanhoth BTW, you can safely merge this since the crash has nothing to do with this PR. It's happening in the develop branch.

stefanhoth:
Thanks for the investigation.

PR #196, opened 2015-02-16
Improvements in ChapterComparator

tasomaniac:
- Instead of keeping a ref to SharedPrefs, ref to HOME GDG is kept.
- MAX_DISTANCE is introduced.
    - When we were at San Francisco at the summit, I saw people having difficulties when they are looking for their home gdg. 
    - With this commit, only GDGs within 500 KM are sorted with the distance
    - Others are sorted alphabetically.

We can change the 500KM max distance but I found that it is enough for most of the countries. I am open to suggestions.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/utils/ChapterComparator.java
+++ b/app/src/main/java/org/gdg/frisbee/android/utils/ChapterComparator.java
@@ -17,10 +17,11 @@ import java.util.Comparator;
  */
 public class ChapterComparator implements Comparator<Chapter> {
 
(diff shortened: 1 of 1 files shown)


friedger:
Maybe have an auto complete field for entering the name for those that know
the name.

Maybe later one can add to search GDGs near the city entered.
Am 16.02.2015 14:29 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
Auto complete will be actually really good. 
But for now, what do you think about this PR?

Issue #163, opened 2014-12-19
Fix GDG Pulse


mauimauer:
This has been broken for quite some time.

tasomaniac:
:arrow_up: :arrow_up: :arrow_up:

HenriqueRocha:
@friedger This can now be closed since you've already merged #197 :+1:

PR #197, opened 2015-02-16
GDG Pulse is fixed by providing the new JSON url.

tasomaniac:
And added some padding to the cards in the Pulse list. Normal padding is not the same. Actually we shouldn't use Cards here. The app is over cardified. Should I remove cards here?


friedger:
Yeah, less cards is more material
Am 16.02.2015 14:46 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
This fixes #163

HenriqueRocha:
@tasomaniac I already an on going branch to remove cards so maybe I should finish removing those no? It will save you some time.

tasomaniac:
The things about cards are out of this PRs scope. If you are already removing them, is there much else to do?

HenriqueRocha:
Yes, I think there is. I'm not focusing on that code now.

tasomaniac:
@HenriqueRocha we can split the job. Lets do it together.

friedger:
A screenshot would be nice

PR #198, opened 2015-02-16
Ripple effect drawn on top of the items with this flag.

tasomaniac:
Some of the lists don't contain touch state but for the ones that do, this fixes ripple effect drawn behind.


tasomaniac:
Here is the screen recordings of the old and new ripple effects. 

broken

fixes

HenriqueRocha:
LGTM :+1:

stefanhoth:
Let's :ship: this! Thanks!

PR #202, opened 2015-02-18
Grade optimisations

tasomaniac:
With shrinkResources, unused images will be removed automatically.
With resConfigs, only the languages that we support will be included from 3rd party libs.
More info at http://cyrilmottier.com/2014/08/26/putting-your-apks-on-diet/

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,6 +35,8 @@ android {
         buildConfigField "String", "GIT_SHA", "\"${gitSha}\""
         buildConfigField "String", "BUILD_TIME", "\"${buildTime}\""
 
+        resConfigs "en", "cs", "de", "es", "fr", "hi", "ko", "lt", "nl", "pl", "pt-rBR", "ru", "sk", "th", "tr", "uk", "zh-rCN", "zh-rHK", "zh-rTW"
+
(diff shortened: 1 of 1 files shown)


tasomaniac:
I got this message during the build:

> Removed unused resources: Binary resource data reduced from 2090KB to 1758KB: Removed 15%

Its about 200KB

HenriqueRocha:
LGTM :+1:

app/build.gradle
```diff
@@ -35,6 +35,8 @@ android {
         buildConfigField "String", "GIT_SHA", "\"${gitSha}\""
         buildConfigField "String", "BUILD_TIME", "\"${buildTime}\""
 
+        resConfigs "en", "cs", "de", "es", "fr", "hi", "ko", "lt", "nl", "pl", "pt-rBR", "ru", "sk", "th", "tr", "uk", "zh-rCN", "zh-rHK", "zh-rTW"
```
stefanhoth: This looks like a potential problem in maintenance as the values-xx folders and this line could diverge. Hope they will figure out a way to have an "auto" in place. From what I see in the bug tracker this is not there yet.

tasomaniac: Yes, I know. As the owners of the projects, when there is a new language contribution, if you remember this config, we will be fine.

Issue #186, opened 2015-02-09
Replace old pull to refresh with new one


HenriqueRocha:

HenriqueRocha:
@stefanhoth Feel free to close this issue once you merge #188

stefanhoth:
@HenriqueRocha Please add a meaningful description the next time, e.g. a link to the updated version, a blog post etc. Makes it easier to understand your task.

PR #203, opened 2015-02-18
Remove NineOldAndroids dependency

HenriqueRocha:
NineOldAndroids was used by the old implementation of the pull to refresh pattern.
With the new implementation in PR #186 this library is not needed anymore.


tasomaniac:
Yey :+1:

stefanhoth:
:clap:

PR #204, opened 2015-02-18
Update OkHttp and Picasso libraries, add Okio

HenriqueRocha:
Update OkHttp from 1.6 to 2.2.0.
Update Picasso to the latest version also.

Add Okio library since according to OkHttp's website:
> You'll also need Okio, which OkHttp uses for fast I/O and resizable buffers.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -107,9 +107,10 @@ dependencies {
 
     //Square
     compile 'com.squareup:otto:1.3.6'
-    compile 'com.squareup.okhttp:okhttp:1.6.0'
-    compile 'com.squareup.okhttp:okhttp-urlconnection:1.6.0'
-    compile 'com.squareup.picasso:picasso:2.3.0'
+    compile 'com.squareup.okhttp:okhttp:2.2.0'
+    compile 'com.squareup.okhttp:okhttp-urlconnection:2.2.0'
+    compile 'com.squareup.okio:okio:1.2.0'
+    compile 'com.squareup.picasso:picasso:2.5.0'
 
     //Jake Wharton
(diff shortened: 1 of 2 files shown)


tasomaniac:
Did it just work? I thought there were API changes in OkHttp. 
Are we actually using OkHttp?

HenriqueRocha:
Yes, it just worked. The app is using the low level API, which apparently wasn't changed, only the high level one. I didn't even get compilation errors. I don't have much time now to work on improving the network layer of the app. I think it's better to focus on the UI stuff now but I'll come back to OkHttp in the future.

tasomaniac:
Yes of course, 
I saw in the previous commits that they upgraded to 2.0 and then revert it back to 1.6. That's why I didn't try it. Thank you.

stefanhoth:
@HenriqueRocha Currently this is not mergeable.

app/build.gradle
```diff
@@ -106,10 +106,11 @@ dependencies {
     compile 'com.android.support:mediarouter-v7:21.0.3'
     compile 'com.google.android.apps.dashclock:dashclock-api:2.0.0'
     compile 'com.squareup:otto:1.3.4'
-    compile 'com.squareup.okhttp:okhttp:1.6.0'
-    compile 'com.squareup.okhttp:okhttp-urlconnection:1.6.0'
+    compile 'com.squareup.okhttp:okhttp:2.2.0'
+    compile 'com.squareup.okhttp:okhttp-urlconnection:2.2.0'
+    compile 'com.squareup.okio:okio:1.2.0'
```
stefanhoth: Do we need to update the open source libraries screen?

PR #206, opened 2015-02-20
Putting Gdg Apk on a diet.

tasomaniac:
ran `image_optim -r app/src/main/res/` for all images and reverted the 9patch images in case there migth be problems with them. It saves around 300K of total size


HenriqueRocha:
Great! :+1:

stefanhoth:
Did you also remove the image properties that make aapt complain? (sRGB profiles or sth like that)

tasomaniac:
I was in my mind but I didn't try something spesific about it. But I hope that it will resolve that as well. 

I've just looked at the build output and saw that the complains are way less then before. And the remaining ones are 9patches. I can look into the 9patch ones as well. They are tricky and if we do automatic anything, they should be loseless.

stefanhoth:
Nevermind, was just curious. Thanks for your effort here.

Issue #174, opened 2015-02-05
Update navigation drawer to Material Design guidelines 


HenriqueRocha:
## Problem
Currently the app uses an old UI for the navigation drawer.

![old_nav_drawer](https://cloud.githubusercontent.com/assets/140122/6305994/715f8d44-b927-11e4-99f4-12a6c349e80d.png)

## Proposed solution
Implement the navigation drawer according to the new Material Design guidelines.
![patterns_navdrawer_settings3](https://cloud.githubusercontent.com/assets/140122/6306007/89282706-b927-11e4-8873-e111088db697.png)

HenriqueRocha:
I'll start working on this issue today on my free time. It might take a bit to implement because I'm moving to London for two weeks but I'll create a PR as soon as it's ready.

friedger:
I think switching user is too much for the GDG app. In the top part of the drawer could go information from the Home GDG.

However, the dropdown in the news activity should remain as @mauimauer stated as well.

HenriqueRocha:
Progress on this issue is going well. I've replaced the ActionBar with the Toolbar everywhere.
I still have to take care of all the styles to make sure that everything looks nice on different versions of Android.

friedger:
@HenriqueRocha Have you seen this https://github.com/mikepenz/MaterialDrawer

HenriqueRocha:
@friedger Not yet, but I'll have a look. Anyway, I've already done so much work in the nav drawer as you can see from my latest PR that it might not be of much help at this point.

tasomaniac:
I saw that library, but I don't like such libraries. It is super easy to make a Nav Drawer with this library but you have to use their implementation and it will be pain in the neck when you want change a look of something. I think we should implement ourselves.

friedger:
Maybe we have the discussion about libs somewhere else :-)

friedger:
Implemented by #253 #252 #247 #244

HenriqueRocha:
This is not done yet. I'm still working on this. :)

tasomaniac:
Great article and sample project about Material Design Drawer. 
https://medium.com/@sotti/navigation-drawer-styling-according-material-design-5306190da08f
https://github.com/Sottti/MaterialDesignNavDrawer

friedger:
I am closing this one and we can add smaller issues for updates of the drawer

tasomaniac:
makes sense.

PR #210, opened 2015-02-22
208/Use light actionable

### LINKED ISSUE #208: GDG Spinner: Background and font have no contrast
Moto X (1st Gen) with Android 4.4.4

image

friedger:
fixes #208 
A quick fix for the dropdown in the action bar

device-2015-02-22-085026


HenriqueRocha:
That will make the icons in the action bar black. That's a temporary solution right?

friedger:
That is not good. But I think changing the theme is the right solution. Needs more work then.

friedger:
@HenriqueRocha I close this and we leave #208 open until #174 has been merged. I assign #174 to you, ok?

PR #212, opened 2015-02-22
211/Use IO extended for special events

friedger:
This PR updates the details for the special events activity. It introduces an object for these details that can be updated in the future through the hub.

In addition, it moves the activity into its own package. It also fixes the SpecialEventActivity's use of end data extra.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -71,7 +71,7 @@
     <activity android:name=".activity.MainActivity" />
     <activity android:name=".activity.GdeActivity" />
     <activity android:name=".activity.AboutActivity" />
-    <activity android:name=".activity.SpecialEventActivity" />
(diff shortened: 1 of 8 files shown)


friedger:
use womentechmakers before io extended

tasomaniac:
Why did you close this? I really liked the SpecialEvents class here. Can I use it and change it to wtm?

friedger:
Yes, please do so. I just closed it because we want wt first.


  [image: --]
Friedger Müffke
<http://about.me/friedger>


2015-02-24 9:50 GMT+01:00 Said Tahsin Dane <notifications@github.com>:

tasomaniac:
I wrote that comment before I saw your branch. That's why we had some
duplicate work but no problem.
You already merged mine. I will send another PR with multiple special event
series support.

PR #213, opened 2015-02-22
184/Use android build gradle plugin 1.1.0 and add simple junit test

friedger:
This PR adds a simple unit tests for EventDateComparator (and also improves the comparator) by using android gradle plugin 1.1.0

screen shot 2015-02-22 at 22 51 54

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -121,6 +121,8 @@ dependencies {
     //Others
     compile 'org.jsoup:jsoup:1.7.3'
     compile 'joda-time:joda-time:2.3'
+    compile 'com.google.code.findbugs:jsr305:2.0.1'
+
     compile ('de.keyboardsurfer.android.widget:crouton:1.8.4') {
         exclude module: 'support-v4'
     }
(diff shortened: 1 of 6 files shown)


tasomaniac:
Looks good. :+1: 

I have a question about findbugs you have just applied. Is it working? I am sure that if it was working, it would have found couple of bugs in the code. :)

app/build.gradle
```diff
@@ -121,6 +121,8 @@ dependencies {
     //Others
     compile 'org.jsoup:jsoup:1.7.3'
     compile 'joda-time:joda-time:2.3'
+    compile 'com.google.code.findbugs:jsr305:2.0.1'
```
HenriqueRocha: Is this relevant since there is no task to actually perform the code quality?

friedger: this is needed as the android gradle plugin uses a version for test and somehow it gets confused without this line.

friedger:
I haven't applied findbugs, it is just to make the android gradle plugin happy. LGTM

HenriqueRocha:
Okay, I'll merge it. I've used it before without findbugs so it's kind of strange.
We can remove it later if we find a workaround.

PR #215, opened 2015-02-24
Special event wtm

tasomaniac:
WTM Special event is added. 

Right now, we have begin and end date in everywhere but we don't use them. Why not?

Instead of updating the app for every event series, there are 2 options I think:

- Using these dates and change the special event dynamically with the time. 
- Have multiple special event support. I like this becuase this was we don't worry about anything. It can also be online. Tags, descriptions and logos can be provided online. It can come from GDG-X Hub or we can even use Google Analytics Tag Manager.


tasomaniac:
I know this is big but if you can merge this, I can continue on implementing multiple special event support.

app/src/main/res/values/strings.xml
```diff
@@ -108,8 +108,10 @@
     <string name="flightschool">Dart Flight School</string>
     <string name="flightschool_description">Dart hit 1.0 and is now Flight Ready! Developer communities around the world are hosting Dart Flight School events during the month of February 2014. Get your web development wings with Dart and Angular!</string>
     <string name="ioextended_description">Google I/O Extended is about joining the Google conference at your local group.</string>
+    <string name="wtm_description">Women Techmakers is Google\'s global program and brand for women in technology. Kicked off in 2012 by then VP of Google[x] Megan Smith, Women Techmakers is now led by Women in Technology Advocate Natalie Villalobos and a global team of Googlers who are passionate about empowering women in technology through increased visibility, community, and resources.\n\nEach of our summits and meetups will be catered to local communities of women in technology. Summits and meetups around the world may feature hands-on coding workshops, inspiring speakers, and networking opportunities.\n\nAll events will occur within the month of March.</string>
```
friedger: This is a bit long (if the screen is small)

app/src/main/java/org/gdg/frisbee/android/special/TaggedEvent.java
```diff
@@ -0,0 +1,97 @@
+package org.gdg.frisbee.android.special;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+
+import org.gdg.frisbee.android.R;
+import org.joda.time.DateTime;
+
+public class TaggedEvent implements Parcelable {
```
friedger: It is more EventSeries or TaggedEventSeries

app/src/main/res/values/strings.xml
tasomaniac: I was thinking to have scrollable description for landscape.

app/src/main/java/org/gdg/frisbee/android/special/TaggedEvent.java
tasomaniac: You are right I will change it

PR #218, opened 2015-02-24
refactor: "plus_id" constant is now in Const file.

tasomaniac:


app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
```diff
@@ -232,14 +237,13 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
     @Override
     public void onDestroy() {
         super.onDestroy();
-        Timber.d("onDestroy()");
```
HenriqueRocha: If you delete the log statement, then you can remove the whole method.

app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
```diff
@@ -64,10 +69,10 @@
 
     private NewsAdapter mAdapter;
 
-    public static NewsFragment newInstance(String plusId) {
+    public static NewsFragment newInstance(final String plusId) {
```
HenriqueRocha: You don't need the final here.

app/src/main/java/org/gdg/frisbee/android/fragment/InfoFragment.java
```diff
@@ -285,15 +286,16 @@ public void onNotFound(String key) {
     }
 
     private String getUrlFromPersonUrl(Person.Urls personUrl) {
+        final String plus_id = getArguments().getString(Const.EXTRA_PLUS_ID);
         if (personUrl.getValue().contains("+")) {
             try {
-                return "+" + URLDecoder.decode(personUrl.getValue().replace("plus.google.com/", "").replace("posts", "").replace("/", "").replace("about", "").replace("u1", "").replace("u0", "").replace("https:", "").replace("http:", "").replace(getArguments().getString("plus_id"), ""), "UTF-8").trim();
+                return "+" + URLDecoder.decode(personUrl.getValue().replace("plus.google.com/", "").replace("posts", "").replace("/", "").replace("about", "").replace("u1", "").replace("u0", "").replace("https:", "").replace("http:", "").replace(plus_id                               , ""), "UTF-8").trim();
```
HenriqueRocha: Why is all the extra space on this line?

app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
tasomaniac: I should. @stefanhoth told me to remove these unnecessary Timber comments.

stefanhoth: Indeed. At least I can't see a benefit of logging the life cycle.

app/src/main/java/org/gdg/frisbee/android/fragment/InfoFragment.java
tasomaniac: It comes from refactoring, fixing now. Thank you.

app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
tasomaniac: Isn't it a good practice to put final on those which shouldn't change.

app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
tasomaniac: I think so too. I removed the function and logs now.

app/src/main/java/org/gdg/frisbee/android/fragment/NewsFragment.java
HenriqueRocha: In C++ yes, not in Java. In Java you can always change the object even if it's final, in C++ you can't, you just can't change the reference, so it's pointless since the reference can only be changed inside the method, nobody else can change it. If the method is big (which shouldn't be) it might be helpful but for small methods it's pointless.

tasomaniac: The benefit is for other future developers I think. If someone tries to make an operation with plusId here, compiler will tell them not to do. But you are right, it is almost pointless for small functions.

tasomaniac:
Fixed the comments from @HenriqueRocha

HenriqueRocha:
LGTM :+1:

PR #219, opened 2015-02-24
fix: Error when arrow drawer item clicked for non login user.

tasomaniac:
Error implementation should be more beautiful.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
@@ -105,7 +105,11 @@ public abstract class GdgNavDrawerActivity extends GdgActivity {
                         navigateTo(PulseActivity.class, null);
                         break;
                     case Const.DRAWER_ARROW:
-                        navigateTo(ArrowActivity.class, null);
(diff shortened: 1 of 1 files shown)


mauimauer:
I'd suggest just hiding the element in the drawer if the user is not logged in.

HenriqueRocha:
I also agree on hiding the element. @tasomaniac Please let me do that since I've already done a lot of work on the navigation drawer and it will be a pain to rebase the code once it's done.

tasomaniac:
I think hiding is not good. It is because people will never know such cool feature exists. 
I think the right thing to do is to open login page with an information about Arrow feature.

HenriqueRocha:
True, showing a login screen would be good.

tasomaniac:
This was really easy to do. I can do that in the future. I will open an
issue for that.
But I think there are lots of things to do for Release 2.0. We can do it
after the release.

HenriqueRocha:
I really think this should be merged. I had to look into the code to understand why nothing was happening when I pressed "Arrow" and I noticed that it was calling `finish()` immediately if the user was not logged in. Very bad UX.

I approve this PR but I'll wait for someone else to comment on this.

tasomaniac:
And the more weird thing was this: I was using the app in landscape, Arrow
page forces it to be portrait and then call finish immediately. :) What
happens is that it changes the orientation on a still tablet without
anything else :)

PR #220, opened 2015-02-24
Selected chapter cannot be null

tasomaniac:
- refactor: EXTRA_GROUP_ID refactored into EXTRA_CHAPTER_ID and removed unused "selected_chapter" key
- `MyAdapter` is now `ChapterFragmentPagerAdapter`
- `initChapters()` function is used everywhere to initialize the `ViewPager`. No more duplicates.
- `initChapters()` now gets selectedChapter as a variable. It is given immediately when an orientation change occurs. It can be null. When it is null, it is gotten from Intent Extra.
- `ChapterFragmentPagerAdapter` now gets selectedChapter from constructor and it cannot be null again.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
@@ -241,8 +241,7 @@ public class FirstStartActivity extends ActionBarActivity implements FirstStartS
 
         Intent resultData = new Intent(FirstStartActivity.this, MainActivity.class);
         resultData.setAction("finish_first_start");
-        resultData.putExtra(Const.EXTRA_GROUP_ID, mSelectedChapter);
-        resultData.putExtra("selected_chapter", mSelectedChapter);
+        resultData.putExtra(Const.EXTRA_CHAPTER_ID, mSelectedChapter);
         startActivity(resultData);
 
         super.finish();
--- a/app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
@@ -20,6 +20,8 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
@@ -54,10 +56,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-
 import butterknife.InjectView;
 import de.keyboardsurfer.android.widget.crouton.Crouton;
 import de.keyboardsurfer.android.widget.crouton.Style;
@@ -77,7 +75,7 @@ public class MainActivity extends GdgNavDrawerActivity implements ActionBar.OnNa
     private Handler mHandler = new Handler();
 
     private ChapterAdapter mSpinnerAdapter;
-    private MyAdapter mViewPagerAdapter;
+    private ChapterFragmentPagerAdapter mViewPagerAdapter;
     private ApiRequest mFetchChaptersTask;
 
     private boolean mFirstStart = false;
@@ -94,7 +92,6 @@ public class MainActivity extends GdgNavDrawerActivity implements ActionBar.OnNa
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        Timber.i("onCreate");
         setContentView(R.layout.activity_main);
 
         GroupDirectory client = new GroupDirectory();
@@ -106,7 +103,6 @@ public class MainActivity extends GdgNavDrawerActivity implements ActionBar.OnNa
         mSlidingTabLayout.setDistributeEvenly(true);
         mSlidingTabLayout.setOnPageChangeListener(this);
 
-        mViewPagerAdapter = new MyAdapter(this, getSupportFragmentManager());
         mSpinnerAdapter = new ChapterAdapter(MainActivity.this, android.R.layout.simple_list_item_1);
         getSupportActionBar().setNavigationMode(ActionBar.NAVIGATION_MODE_LIST);
         getSupportActionBar().setListNavigationCallbacks(mSpinnerAdapter, MainActivity.this);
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
```diff
@@ -241,8 +241,7 @@ public void finish() {
 
         Intent resultData = new Intent(FirstStartActivity.this, MainActivity.class);
         resultData.setAction("finish_first_start");
-        resultData.putExtra(Const.EXTRA_GROUP_ID, mSelectedChapter);
-        resultData.putExtra("selected_chapter", mSelectedChapter);
```
tasomaniac: this extra was not used.

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -77,7 +75,7 @@
     private Handler mHandler = new Handler();
 
     private ChapterAdapter mSpinnerAdapter;
-    private MyAdapter mViewPagerAdapter;
```
tasomaniac: Refactoring of class name with "My" :)

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -106,7 +103,6 @@ public void onCreate(Bundle savedInstanceState) {
         mSlidingTabLayout.setDistributeEvenly(true);
         mSlidingTabLayout.setOnPageChangeListener(this);
 
-        mViewPagerAdapter = new MyAdapter(this, getSupportFragmentManager());
```
tasomaniac: Initializing the adapter with nothing does nothing. It is initialized in only 1 place now which is `initChapters`

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -163,18 +159,9 @@ public void onNotFound(String key) {
...
                 ArrayList<Chapter> chapters = savedInstanceState.getParcelableArrayList("chapters");
-                mSpinnerAdapter.clear();
-                mSpinnerAdapter.addAll(chapters);
-
-                if (savedInstanceState.containsKey("selected_chapter")) {
-                    Chapter selectedChapter = savedInstanceState.getParcelable("selected_chapter");
-                    selectChapter(selectedChapter);
-                } else {
-                    mViewPagerAdapter.setSelectedChapter(chapters.get(0));
-                }
```
tasomaniac: These are gone and the same thing done in `initChapters`. No more duplicate ftw

tasomaniac:
I think this should be merged after #218

HenriqueRocha:
Okay, I'll have that into account.

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -203,31 +190,39 @@ protected void onResume() {
...
-            }
-        } else {
-            chapter = chapters.get(0);
         }
+        
+        if (selectedChapter == null) {
+            selectedChapter = chapters.get(0);
+        }
+
+        getSupportActionBar().setSelectedNavigationItem(mSpinnerAdapter.getPosition(selectedChapter));
```
HenriqueRocha: Nooooo!!! That doesn't exist anymore! I'll have to rebase that! :crying_cat_face: 
That method is deprecated and I've changed that since we won't be using an action bar anymore.
But it's okay, I'll deal with it.

tasomaniac: I was pretty certain about this will make conflicts. 
I can rebase this onto your commit when you are done. This PR is not urgent.

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -203,31 +190,39 @@ protected void onResume() {
...
+     * 
+     * If the selectedChapter is null, tries to get it from Intent Extras,
+     *                                  if not found, gets the first Chapter.
+     * This function is called after cache query or network call is success,
+     * or after an orientation change using savedInstance.
+     *  
+     * @param chapters Chapter array to be initialized, never null.
+     * @param selectedChapter selectedChapter, can be null.
+     */
+    private void initChapters(@NonNull ArrayList<Chapter> chapters,
```
HenriqueRocha: Please change this method into two methods. If you are doing code review and you read the following `initChapters(chapters, null);` and before it was just `initChapters(chapters);` you will wonder what that `null` argument is and you have to check the new signature.

It's better to have the method just do one thing instead of two (one for null, one for not null) so that it's obvious what's happening just by reading the line.

tasomaniac: It was in my mind and I missed that. You are absolutely right. Updating...

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -314,13 +311,14 @@ public boolean onNavigationItemSelected(int position, long l) {
...
-    public class MyAdapter extends FragmentStatePagerAdapter {
+    public class ChapterFragmentPagerAdapter extends FragmentStatePagerAdapter {
         private Context mContext;
         private Chapter mSelectedChapter;
 
-        public MyAdapter(Context ctx, FragmentManager fm) {
+        public ChapterFragmentPagerAdapter(Context ctx, FragmentManager fm, @NonNull Chapter selectedChapter) {
             super(fm);
             mContext = ctx;
+            this.mSelectedChapter = selectedChapter;
```
HenriqueRocha: When you use the `m` prefix in a field name, you don't use the `this.`

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -364,11 +362,10 @@ public Chapter getSelectedChapter() {
             return mSelectedChapter;
         }
 
-        public void setSelectedChapter(Chapter chapter) {
+        public void setSelectedChapter(@NonNull Chapter chapter) {
```
HenriqueRocha: If you add `@NonNull` here than it's not worth checking the `null` again inside. You should safely assume that the client of the method is fulfilling the non null pre-condition. The IDE will tell the client of the method that a null argument should not be passed. So, either you remove `@NonNull` or remove the check for `!= null` inside.

HenriqueRocha: Sorry, you're checking `chapter` and not `mSelectedChapter`. I misread. Forget about my previous comment.

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
tasomaniac: How about now? It is now really better.

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
tasomaniac: Yep, actually I didn't change anything. I've just put NonNull, becuase it shouldn't be null. 
Actually we can remove that null check as well, but I don't want to mess with analytics code.

PR #221, opened 2015-02-24
minSdkVersion is now 16.

### LINKED ISSUE #217: Drop support for SDK older than 16
While implementing the new navigation drawer and getting more familiar with the code base I noticed many places where we're using code which requires API 11 or 14, even tough we're supporting API 10.

Due to the nature of this app, maybe it would be interesting to see the usage distribution of active users among the different APIs and maybe drop support for everything below 16 in the next version.

I don't know who has access to the Developer Console to check this out. Maybe @mauimauer ?

tasomaniac:
All the device version checks below 16 are removed.

This should also close #217


HenriqueRocha:
LGTM :+1:

tasomaniac:
Oh you can merge now :)

friedger:
You too
Am 24.02.2015 15:39 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
No I cannot :(

tasomaniac:
Yesssss !!!

PR #223, opened 2015-02-24
Joda Time for Android by dlew is used instead of the original.

tasomaniac:
They say that this is better in terms of performance.
It is because it handles String resources with Android way with resources.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -120,7 +120,7 @@ dependencies {
 
     //Others
     compile 'org.jsoup:jsoup:1.7.3'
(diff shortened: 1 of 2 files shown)


HenriqueRocha:
@tasomaniac This is not building. Is it building on your machine?

tasomaniac:
It was and I was surprised. I think it was because I didn't clean. Let me check again.

tasomaniac:
It was because of the release build with proguard. I've added a proguard configuration. Lets see now. I will also try the release build on a real device and update the thread.

PR #225, opened 2015-02-25
Lint should abort on error.

tasomaniac:
If there are errors, those shall be fixed. :P


HenriqueRocha:
:-1: I don't believe that Lint should be a blocker even tough the errors should be fixed.

friedger:
I agree with Henrique

friedger:
But we can update this pull request from time to time and see where lint breaks! Like here

HenriqueRocha:
@tasomaniac Do you want to work on fixing the Lint errors? It would probably be a good idea while I'm rewriting the navigation drawer.

tasomaniac:
Yep I can work on that and create a separate PR and we can keep this PR.

stefanhoth:
I approve. I know it's frustating to have a build fail on lint but that's what it's for, to show you where you did bad things and failing the build before the merge is better than have other people fixing it some time later - if at all.

But I would say that we should have a look at the list of "fatal" lint errors and curate this first before we force everyone to work like that.

[See gradle docs on lint](http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Lint-support) with 

```
        // Set the severity of the given issues to fatal (which means they will be
        // checked during release builds (even if the lint target is not included)
        fatal 'NewApi', 'InlineApi'
        // Set the severity of the given issues to error
        error 'Wakelock', 'TextViewEdits'
        // Set the severity of the given issues to warning
        warning 'ResourceAsColor'
        // Set the severity of the given issues to ignore (same as disabling the check)
        ignore 'TypographyQuotes'
```

tasomaniac:
Today I fixed a string-array resource problem for example. There were 6 items in the English version but there were 5 in the other languages. These kind of things actually crash the app for other languages and you never notice. 
There are lots of Lint warnings that we can make the build pass. But for the errors, I personally think that it should break the build.

friedger:
This issue is not about a general decision but about the current status of the code. Everyone is invited to make this build pass. But everyone is also invited to work on other issues.

stefanhoth:
```
Check performed at Tue Mar 03 23:15:59 CET 2015.
141 errors and 344 warnings found:
```

Wow, much work. I guess before we are in a better state we can't merge this PR as it would break all other PRs.

stefanhoth:
@tasomaniac Where do you have your fixes? Care to push them here, too?

friedger:
easiest is to merge develop


  [image: --]
Friedger Müffke
<http://about.me/friedger>


2015-03-03 23:30 GMT+01:00 Stefan Hoth <notifications@github.com>:

tasomaniac:
@stefanhoth it is already merged #266

stefanhoth:
Ok, thanks. Current status:

```
Check performed at Tue Mar 03 23:45:07 CET 2015.
131 errors and 324 warnings found:
```

tasomaniac:
Even though some of the errors are converted to warnings, the build is a success now.

Issue #211, opened 2015-02-22
Use IO Extended as special events


friedger:
The special events should be updated for IO extended 2015 events

tasomaniac:
Or Android Study Jams? Or Women Techmakers? We can do extended after study
Jams.

friedger:
In the long term, we just change the data on the hub.

Is there WTM happening globally? What is the time frame?

I can imagine to promote events on the main activity/chapter page either on
a regional level or chapter level. Again this should probably happen
through the hub.

As 2.0 will be hopefully released in mid/end march, study jam has partly
finished and it is 8 weeks before io extended, good enough to start
promoting it...




  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-02-23 6:43 GMT+01:00 Said Tahsin Dane <notifications@github.com>:

tasomaniac:
If we release the 2.0 at that time, it will be too late for Study Jams. WTM
events will happen during March. And yes it is global. Just like DevFest's
but not that big.

friedger:
So lets get a version ready for tag "wtm", probably best a branch from the currently released version, only with the special events updated.

friedger:
Super powers ☺
Am 25.02.2015 13:00 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
:smile: :battery:

PR #226, opened 2015-02-25
Multiple event series support

tasomaniac:
Basically event series are generated and stored when App open in `App` class.
They are added only if the device is on time for that event. 

Questions:
- Right now, after June 1st, there will not be special event in the drawer. Should that be the case? Right now when we open the current production app, there is DevFest and the list shows unrelated "devfest" tagged events. Which is the right way?
- Should we have some online functionality to do this? Like coming from hub? Hub can say that these are the current GDG event series that are going on throughout the world. 

Screenshots:
- **As of now**
screenshot_2015-02-25-11-12-52
- **on April 1st**
screenshot_2015-04-02-11-14-12
- **on May 1st**
screenshot_2015-05-01-11-14-41
- **on June 1st**
screenshot_2015-06-01-11-30-32


tasomaniac:
@HenriqueRocha I know that this will cause conflicts with your work. Just review it and I can rebase this onto yours when you are ready.

app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
```diff
@@ -37,14 +37,17 @@
...
+
+import java.util.ArrayList;
 
 import butterknife.InjectView;
+import butterknife.OnItemClick;
 import de.keyboardsurfer.android.widget.crouton.Crouton;
 import de.keyboardsurfer.android.widget.crouton.Style;
 
-public abstract class GdgNavDrawerActivity extends GdgActivity {
+public abstract class GdgNavDrawerActivity extends GdgActivity{
```
HenriqueRocha: Why was the space before the brace removed?

app/src/main/java/org/gdg/frisbee/android/Const.java
```diff
@@ -97,4 +96,12 @@
 
     public static final String TRUSTSTORE_PW = "VU%&ibkr45pnq39v53x";
     public static final boolean ALPHA = false;
+
+    public static final long START_TIME_IOEXTENDED = 1430438400000L;
+    public static final long START_TIME_STUDY_JAMS = 1422748800000L;
+    public static final long START_TIME_WTM = 1422748800000L;
```
friedger: would be nice to know what the dates are for humans

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -127,9 +131,46 @@ public void onCreate() {
         mLocationFinder = new GingerbreadLastLocationFinder(this);
         mLocationFinder.setChangedLocationListener(this);
         updateLastLocation();
+        
+        //Init TaggedEventSeries
+        mTaggedEventSeriesList = new ArrayList<>();
+        addTaggedEventSeriesIfDateFits(new TaggedEventSeries("wtm",
+                R.drawable.drw_ic_wtm,
```
friedger: while wtm is the tag, I think women techmakers should be abbreviates as wt, but that is not an issue here.

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -127,9 +131,46 @@ public void onCreate() {
         mLocationFinder = new GingerbreadLastLocationFinder(this);
         mLocationFinder.setChangedLocationListener(this);
         updateLastLocation();
+        
+        //Init TaggedEventSeries
```
HenriqueRocha: Please remove the comment and put the corresponding code in a `initTaggedEventSeries()` method.

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -127,9 +131,46 @@ public void onCreate() {
...
+                R.drawable.drw_ic_ioextended,
+                R.string.ioextended,
+                R.string.ioextended_description,
+                R.drawable.ic_ioextended,
+                Const.START_TIME_IOEXTENDED,
+                Const.END_TIME_IOEXTENDED));
     }
 
-    public void migrate(int oldVersion, int newVersion) {
+    /**
```
HenriqueRocha: This kind of Javadoc is useless because it's very redundant with the code, specially since it's a private method. You can remove it.

friedger:
Looks good to me, @HenriqueRocha should merge when he feels fits best.

Yes, finally the data come from the hub! Any ambitions to do that? Main point is to create the web ui

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -245,4 +286,15 @@ public boolean isOrganizer() {
...
     }
+
+    /**
+     * Return the current list of GDG event series occurring in the world.
+     * This may be empty but cannot be null.
+     *
+     * @return Array of current event series. 
+     */
+    @NonNull
+    public ArrayList<TaggedEventSeries> getTaggedEventSeriesList() {
```
HenriqueRocha: Remove Javadoc and rename method to `currentEventsOccurring()`
The problem with comments is that it's hard to make them correct or in sync with the code, for example, you wrote "The current list" and in the `@return` you say "Array of".

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
```diff
@@ -18,22 +17,22 @@
...
+                             @DrawableRes int logoResId,
+                             long startDateInMillis,
+                             long endDateInMillis) {
         mTag = tag;
         mDrawerIconResId = drawerIconResId;
         mTitleResId = titleResId;
         mDescriptionResId = descriptionResId;
         mLogoResId = logoResId;
-        mSartDateInMillis = DateTime.now().getMillis();
+        mSartDateInMillis = startDateInMillis;
```
HenriqueRocha: Just a small typo: `mSartDateInMillis` should be renamed to `mStartDateInMillis`.

HenriqueRocha:
Apart from what I mentioned, LGTM. Once you fix those things I'll merge it and rebase my branch and fix all the conflicts.

app/src/main/java/org/gdg/frisbee/android/Const.java
tasomaniac: I did it like this because it was like this before. Changing now.

app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
tasomaniac: Good catch. I was working on checkstyle integration. It is almost finished but I will finish it after Release 2.0

app/src/main/java/org/gdg/frisbee/android/app/App.java
tasomaniac: The official tag is "wtm" :)

app/src/main/java/org/gdg/frisbee/android/app/App.java
tasomaniac: Some say that every function should be javaDoc ed. And I personally like that idea. Does it harm anything?

What are the comments about this? Actually it is important because when I integrate checkstyle, I should make the configuration appropriately. The default checkstyle config is to have javaDocs on all functions.

HenriqueRocha: According to the Clean Code book by Robert Martin, you should only add Javadoc if it's a public API. If you haven't read that book, I strongly recommend that you start reading it yesterday! ;)
Please configure Checkstyle to ignore missing Javadocs.
Usually, Javadoc on private methods can be avoided and most of the time is completely redundant.

tasomaniac: Makes sense. I didn't read it and in fact I already started yesterday. :P

Actually for the initial checkstyle integration, I will remove things about JavaDocs completely becuase it is really hard to integrate it with the whole codebase. :)

tasomaniac:
This also closes #211

friedger:
Nice work!
Am 25.02.2015 13:00 schrieb "Henrique Rocha" <notifications@github.com>:

PR #228, opened 2015-02-26
Small crash fix.

tasomaniac:


tasomaniac:
Sorry guys, it is a really small fix and I have to merge this to continue.

friedger:
If it does not interfere with other work it is ok.

PR #229, opened 2015-02-26
Removed unnecessary below API-11 resources

tasomaniac:


HenriqueRocha:
LGTM :+1:

app/src/main/res/values/gdg_styles.xml
```diff
@@ -30,20 +30,16 @@
     <item name="android:button">@drawable/btn_check_holo_light</item>
```
friedger: There is more holo

PR #231, opened 2015-02-26
Keys are moved to local.properties file.

tasomaniac:
It is already ignored in the default project configuration. It won't be pushed to github accidentially.

`play_app_id` resource string is trailed with double quotes because of an Android bug described below. 
More info about the bug: https://github.com/jamesward/AndroidStringBug


tasomaniac:
BTW keys should be written in local.properties as key value pairs like below 

play_app_id=1234567890

PR #235, opened 2015-02-26
Chore/Remove meta data from images

friedger:
I just applied exiftool to all pngs to get rid of the warning about unrecognized profiles


tasomaniac:
I did a similar thing with pngout but I was not sure about 9patches and reverted them. We are sure that there is no visiual changes right?

Issue #245, opened 2015-02-28
Remove duplicate code from InfoFragment


friedger:
InfoFragment uses the same code as GgdNavDrawerActivity to retrieve chapter information. InfoFragment should use the gplus client of the activity.

friedger:
implemented in #251

Issue #246, opened 2015-02-28
Improve access to Preferences


friedger:
Maybe remove "gdg" preferences and use DefaultSharedPreferences

or create an GDGPreferenceObject that wraps the "gdg" preferences

mauimauer:
But beware: you might want to have a proper migrate function (ie. on versionCode jump, move settings out of the old pref instance) or this might screw things up

friedger:
Do you have a motivation for using "gdg" instead of the default ones?
I don't care if it is only in on place defined.

tasomaniac:
I think we can continue using "gdg" as the preferences. The problem was I
think manually writing "gdg" string everytime we use. If we can have a
utility class that controls sharwdprefs, isn't it enough?

friedger:
I don't like that we have to put the context in all static methods.

mauimauer:
why not just extract the "gdg" string to our static constants file.

PR #251, opened 2015-03-01
Feature/Handle change of home

friedger:
This PR handles the change of the Home chapter by the user, i.e. the MainActivity and the NavigationDrawer is updated (see #245)

Also preferences are moved to PrefUtils.


friedger:
This is not #246 , but moves preferences closer together.

app/src/main/java/org/gdg/frisbee/android/Const.java
```diff
@@ -26,21 +26,6 @@
  */
 public class Const {
 
-    // Settings
-    public static final String SETTINGS_FIRST_START = "gdg_first_start";
```
friedger: all preference keys are no in PrefUtils

app/src/main/java/org/gdg/frisbee/android/utils/PrefUtils.java
```diff
@@ -3,37 +3,150 @@
...
 import android.content.SharedPreferences;
 
+import org.joda.time.DateTime;
+
 public final class PrefUtils {
-    private static final String PREF_NAME = "gdg";
-    private static final String PREF_OPEN_DRAWER_ON_START = "open_drawer_on_start";
-    private static final String PREF_SIGNED_IN = "gdg_signed_in";
+    public static final String PREF_NAME = "gdg";
+    public static final String SETTINGS_HOME_GDG = "gdg_home";
```
friedger: preferences that can be changed by the user are prefixed with SETTINGS

app/src/main/java/org/gdg/frisbee/android/utils/PrefUtils.java
```diff
@@ -3,37 +3,150 @@
...
+    private static final String PREFS_SEASONS_GREETINGS = "seasons_greetings";
+
+    private static final boolean PREFS_FIRST_START_DEFAULT = true;
 
     private PrefUtils() {
         // Prevent instances of this class being created.
     }
 
+    public static SharedPreferences prefs(final Context context) {
+        return context.getSharedPreferences(PREF_NAME, Context.MODE_PRIVATE);
```
friedger: Maybe we can create an object that wraps this SharedPreferences and gives a nice API to access the preferences. In PrefUtils there are too many static methods.

friedger:
App crashes as described in #255

friedger:
App does not crash anymore. However, the spinner and fragments are not updated.

friedger:
My head is hurting. There are too many options that can happen.  I have the vage feeling that it is working now. LGTM!

app/src/main/java/org/gdg/frisbee/android/Const.java
tasomaniac: makes sense.

app/src/main/java/org/gdg/frisbee/android/adapter/ChapterAdapter.java
```diff
@@ -76,4 +76,14 @@ public long getItemId(int position) {
         else
             return 0;
     }
+
+    public Chapter findById(final String chapterId) {
+        for (int i = 0; i < getCount(); i++) {
```
tasomaniac: I was thinking about this. 
Since `equals` override in Chapter uses GplusId, we can use `indexOf` function in the ArrayList stored in this adapter.

I would be more elegant? Maybe not? What do you think?

app/src/main/java/org/gdg/frisbee/android/utils/ChapterComparator.java
```diff
@@ -18,10 +18,11 @@
 public class ChapterComparator implements Comparator<Chapter> {
 
     private static final float MAX_DISTANCE = 500000;
-    private final String mHomeGdg;
+    @Nullable
+    private final String mHomeChapterId;
 
-    public ChapterComparator(SharedPreferences prefs) {
-        mHomeGdg = prefs.getString(Const.SETTINGS_HOME_GDG, "");
+    public ChapterComparator(String homeChapterId) {
```
tasomaniac: Then, we can add @Nullable annotation to the constructer as well

tasomaniac:
:+1: looks good.

HenriqueRocha:
Really good work! :+1:

app/src/main/java/org/gdg/frisbee/android/adapter/ChapterAdapter.java
friedger: The list is not exposed to the outside, so I don't see a possiblity here. getAll() iterates as well through all elements.

PR #260, opened 2015-03-01
Use blue background and white font in first activity

friedger:
Just a simple change of colors. How do you like it? 

Button and check boxes could do with a better theming, but that is a different PR.

device-2015-03-01-222559
device-2015-03-01-222643
device-2015-03-01-222716

### DIFF
--- a/app/src/main/res/layout-land/fragment_welcome_step1.xml
+++ b/app/src/main/res/layout-land/fragment_welcome_step1.xml
@@ -3,7 +3,7 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
   android:layout_width="match_parent"
   android:layout_height="match_parent"
-  android:background="@color/cloud"
+  android:background="@color/theme_primary"
   android:orientation="vertical">
 
   <LinearLayout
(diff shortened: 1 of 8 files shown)


HenriqueRocha:
The icons look really weird for me. The grey and the blue doesn't match.
I think you should work on this and do a complete PR once everything looks nice.

friedger:
@HenriqueRocha Are you working on theming the widgets?

tasomaniac:
If this is only because of the white spinner, let don't do this.
It looks really weird to me now. The world image should be changed in the
first screen. We can then decide. But we are not designers, it wouldn't be
easy I think :)

HenriqueRocha:
Yes, if it's just because of the white text on the spinner I can change that since that's happening because I changed the ChapterAdapter. I didn't realize it was being used in the start flow.

HenriqueRocha:
@friedger No, right now I'm just redoing the implementation of the drawer.
It's not a `ListView` anymore and doesn't use the `DrawerAdapter`.
It's a `LinearLayout` inside a `ScrimInsetsScrollView`.
I'll create a PR as soon as I have something good enough to be worked on top of to avoid people from adding stuff the `DrawerAdapter` which is not used anymore.

friedger:
I like to have more blue in the app, but I am fine with not having it merged for 2.0

mauimauer:
sorry, but this doesn't look right at all. Also I don't think we should fill-blue everything. Keep white as a contrast.

Issue #261, opened 2015-03-01
Event activity has black text in toolbar


friedger:
text and icons shoud be white

**Current problem:**
![device-2015-03-01-224055](https://cloud.githubusercontent.com/assets/1449049/6433115/61afd39a-c066-11e4-9cdf-5002ac0535e0.png)

friedger:
implemented in #262

PR #262, opened 2015-03-01
use toolbar for EventActivity, use color references were possible

friedger:
Use a white text and blue popup as well

**Solution**
device-2015-03-01-233503

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -145,10 +145,8 @@
     </activity>
     <activity
       android:name=".event.EventActivity"
+      android:theme="@style/GDG.NoActionBar"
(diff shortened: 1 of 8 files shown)


friedger:
@HenriqueRocha I just realized now that I haven't assigned me to #261

HenriqueRocha:
Why blue pop up? It's usually white with black text. If we do this here we would have to change it everywhere. Maybe we should post-pone changing the pop colour for now no?

friedger:
I couldn't find any other overflow menu in the app. So this would be the
only plae


  [image: --]
Friedger Müffke
<http://about.me/friedger>


2015-03-01 23:43 GMT+01:00 Henrique Rocha <notifications@github.com>:

HenriqueRocha:
:+1:

PR #266, opened 2015-03-02
about_tabs are removed on other languages.

tasomaniac:
This is one of the lint errors. about_tabs should contain 6 items but all of these include 5.


friedger:
Good find!

PR #267, opened 2015-03-02
fix: News Fragment errors are fixed.

tasomaniac:
- I now use RecyclerView for StaggeredList. 
- MyView is removed. 
- Refresh animation for loading is fixed. 
- UI is exactly the same. 
- When there are EmptyText, or loading widgets with RecyclerView, SwipeRefreshLayout causes problems with the scroll.
- ContextMenu wasn't implemented. I can do it it if you want to.


app/src/main/java/org/gdg/frisbee/android/adapter/NewsAdapter.java
```diff
@@ -318,18 +318,26 @@ private void populateArticle(ViewHolder mViewHolder, ViewGroup container, final
...
             mViewHolder.articleImage.setVisibility(View.GONE);
-        else {
-            String imageUrl = attachment.getImage().getUrl();
-            if(attachment.getFullImage() != null)
+        } else {
+            String imageUrl = null;
+            if (attachment.getFullImage() != null) {
                 imageUrl = attachment.getFullImage().getUrl();
+            }
+            if (attachment.getImage() != null) {
```
friedger: should be an else if?

tasomaniac: Yes definitely. Good catch

app/src/main/java/org/gdg/frisbee/android/fragment/GdgRecyclerFragment.java
```diff
@@ -0,0 +1,324 @@
...
+import android.view.animation.Animation;
+import android.view.animation.AnimationUtils;
+
+import org.gdg.frisbee.android.R;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+
+
+public class GdgRecyclerFragment extends Fragment {
```
tasomaniac: This is the exact same as GdgListFragment except this uses RecyclerView

PR #268, opened 2015-03-02
chore: remove Joda-Time from 3rd party libraries

HenriqueRocha:
Joda-Time was removed but it was left in the list of 3rd party libraries.


tasomaniac:
It is weird, Joda Time Android is basically a copy of Joda Time with Android in mind. Shouldn't we mention Joda Time as well?

HenriqueRocha:
The mention should be transitive in this case. We use Joda Time Android so that's the one we mention. Joda Time Android will reference the original Joda Time.

PR #275, opened 2015-03-03
Checkstyle integration

tasomaniac:
Remaining things:
- JavaDoc enforcement. We can force it for public functions or classes. It will be too hard for initial integration. 
- MultipleStringLiterals. It is the things like "gdg" Strings. It doesn't let us do that. It forces to move them in Constants file. We can do that later. 
- White space around 
   - The problem was the line 130 at https://github.com/gdg-x/frisbee/blob/develop/app/src/main/java/org/gdg/frisbee/android/adapter/PulseAdapter.java
    - Whats up with `*-` Checkstyle doesn't let us do that.


tasomaniac:
From now on, code will not pass Travis CI if the style is wrong.

PR #277, opened 2015-03-03
Special event default icon

tasomaniac:
This started as a fix and become kind a feature :)

This PR includes the following:

- Default Icon support for TaggedEventSeries
- Theme attributes for TaggedEventSeries values. 
- Theme style is used instead of ugly constructor. 

screenshot_2015-03-03-16-15-19


app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
```diff
@@ -59,11 +60,24 @@
...
 
     public EventAdapter(Context ctx) {
+        this(ctx, R.drawable.icon);
+    }
+    
+    public EventAdapter(Context ctx, @DrawableRes int defaultIcon) {
         mContext = ctx;
         mInflater = LayoutInflater.from(mContext);
         mEvents = new ArrayList<>();
+        mDeafultIcon = defaultIcon;
```
friedger: typo here

app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
```diff
@@ -59,11 +60,24 @@
...
+    }
+
+    /**
+     * The default icon is used if there is no icon for an event.
+     */
+    public void setDeafultIcon(@DrawableRes int mDeafultIcon) {
```
friedger: and here

app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
```diff
@@ -59,11 +60,24 @@
     private Context mContext;
     private LayoutInflater mInflater;
     private ArrayList<Item> mEvents;
+    private int mDeafultIcon;
 
     public EventAdapter(Context ctx) {
+        this(ctx, R.drawable.icon);
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
friedger: do we need a setter? better make it final and developers need to create a new adapter when showing a new tag

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
```diff
@@ -14,25 +19,53 @@
...
     private DateTime mEndDateInMillis;
 
-    public TaggedEventSeries(String tag, 
+    public TaggedEventSeries(Context context,
+                             @StyleRes int specialEventTheme,
+                             @NonNull String tag,
+                             DateTime startDateInMillis,
+                             DateTime endDateInMillis) {
+
+        final ContextThemeWrapper themeWrapper = new ContextThemeWrapper(context, specialEventTheme);
```
friedger: This only works for programmed themes, correct? If we get the data from the server we don't need this. Hence it might make more sense to wrap the ids in a java object and create this either from the programmed theme or from server data.

friedger:
nice!

app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
tasomaniac: I was going to do like that but it is instantiated and set in
EventListFragment rather then in TaggedEventSeriesFragment. So I didn't
want to change that.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
tasomaniac: I also thought about this. There is no easy way to change the theme of an
activity without having a theme as a style.
Overscroll color, button colors actionbar background. All of these have to
be updated one by one programmatically. At least that's the result of my
research. Please fix me if you know how to.

Also the icons colors texts will not come from resources. They will come
from the API.

So if that is the case, we have to change lots of things here for dynamic
event series.

What do you think?

app/src/main/java/org/gdg/frisbee/android/adapter/EventAdapter.java
tasomaniac: I've just changed it. It is now final.

PR #280, opened 2015-03-03
[URGENT] Crash fix: ListView header set before adapter is set.

tasomaniac:
It is fine in Lollipop because you can set headers in ListViews in Lollipop whenever you want.
But before lollipop, the header should be set before Adapter initialisation. So the TaggedEventSeries is crashing for all devices below Lollipop. Should we send an update to the play store?


friedger:
Looks good.

PR #284, opened 2015-03-03
IO Extended should be 2015

tasomaniac:
And translatable="false" added to the event names and the app name.


friedger:
does crowdin understand translatable?

tasomaniac:
I hope so :)

PR #299, opened 2015-03-06
My proposal to fix #298

### LINKED ISSUE #298: App crashes when error occurs in first step
via Crashalytics:

java.lang.IllegalStateException: Fragment FirstStartStep1Fragment{42843128} not attached to Activity
       at android.support.v4.app.Fragment.getResources(Fragment.java:619)
       at android.support.v4.app.Fragment.getString(Fragment.java:641)
       at org.gdg.frisbee.android.fragment.FirstStartStep1Fragment$2.onErrorResponse(FirstStartStep1Fragment.java:113)
       at com.android.volley.Request.deliverError(Request.java:517)
       at com.android.volley.ExecutorDelivery$ResponseDeliveryRunnable.run(ExecutorDelivery.java:101)
       at android.os.Handler.handleCallback(Handler.java:733)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:146)
       at android.app.ActivityThread.main(ActivityThread.java:5598)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:515)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1283)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1099)
       at dalvik.system.NativeStart.main(NativeStart.java)

tasomaniac:
The old code with the crash #298 is like this

```
                if (isDetached()) {
	                 Toast.makeText(getActivity(), R.string.fetch_chapters_failed, Toast.LENGTH_SHORT).show();
                } else {		
                    Crouton.makeText(getActivity(), getString(R.string.fetch_chapters_failed), Style.ALERT).show();		
                 }
```

Crash happens with the `getString` function there. In order that to happen, `isDetached()` should be false meaning that the Fragment is Attached. But the crash says that the fragment is not attached. Damn Fragments!!!

I don't know a permanent solution to this. It is a pain. Whatever you do, you will always have these issues. And they are almost always happen when the user exits the app, without waiting the response (which is obviously fine). 

I would wrap with `try-catch`s all the Crouton's

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/GdeActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/GdeActivity.java
@@ -79,7 +79,10 @@ public class GdeActivity extends GdgNavDrawerActivity {
         }, new Response.ErrorListener() {
             @Override
             public void onErrorResponse(VolleyError volleyError) {
-                Crouton.makeText(GdeActivity.this, getString(R.string.fetch_gde_failed), Style.ALERT).show();
+                try {
+                    Crouton.makeText(GdeActivity.this, R.string.fetch_gde_failed, Style.ALERT).show();
+                } catch (IllegalStateException exception) {
+                }
                 Timber.e("Could'nt fetch GDE list", volleyError);
             }
         });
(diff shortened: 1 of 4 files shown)


friedger:
It should be 
`if (isAttached) showSomething else //don't do anything`

tasomaniac:
Are you sure about that? I don't trust it. If you trust I can change it.

friedger:
isAttached is something else than not isDetached. Hence, I would trust it.

Issue #301, opened 2015-03-07
Youtube video crash


tasomaniac:
Here is the screenshot attached about the crash. It happened when I clicked a video in the news feed page

![screenshot_2015-03-07-23-45-37](https://cloud.githubusercontent.com/assets/763339/6541538/0199a7e0-c4e1-11e4-865b-57769c60f976.png)

tasomaniac:
closed via #304

PR #304, opened 2015-03-11
Fix 301/App crashes when clicking on video

friedger:
This PR handles edge cases where the video url is not in the expected format and there is no app to handle the link.


tasomaniac:
This fixes #301 right?

friedger:
Yes

PR #306, opened 2015-03-13
Joyful Retrofit integration

### LINKED ISSUE #227: Joy/Use retrofit for hub api calls

tasomaniac:
The project now uses Retrofit everywhere, Volley is removed. 

I open the PR early to be reviewed. 

Things that are problematic:
- GCM Registration is not working anymore. I am sending an header like this: `Authorization: Bearer some_long_token` but it returns `401` invalid token. It seems that Volley was sending the same way. 
- We should look into the removed classes and see if we miss header or something to send extra. 

Future work:
- Remove old implementation of Google Plus API. And try to use the APIs from Google Play Services. 
```
    compile 'com.google.http-client:google-http-client-gson:1.19.0'
    compile 'com.google.apis:google-api-services-plus:v1-rev126-1.18.0-rc'
```

This PR fixes #227


tasomaniac:
@mauimauer you are the one who knows the best about the auth and headers to use with Hub API. Can you look and tell the points we should be careful.

stefanhoth:
:+1: Awesome PR. I will review it later if noone else does it before me.

tasomaniac:
Really :) we have to be careful about this. I've changed lots of code here.
I will double check.

Have anyone checked or tested?

PR #307, opened 2015-03-18
fix: proper navigation icon and title

PareshMayani:
- Title in the Settings screen was missing
- Navigation icon was used in About and Settings screen, rather it should use up icon when an activity is a standalone activity.

After Fix:

screenshot_2015-03-19-10-23-35


tasomaniac:
:+1:

tasomaniac:
We should also fix the paddings in the settings page. The icons are not aligned. Settings title and the other titles are not aligned.

PR #308, opened 2015-03-18
fix: padding issue

PareshMayani:
1) Fix: padding issue in details box.

screenshot_2015-03-18-15-50-08

2) Fix: padding issue in change log details textview
screenshot_2015-03-18-16-00-31

3) Fix: padding issue in get involved details layout
screenshot_2015-03-18-16-00-34


tasomaniac:
Nice, the landscape of the first info screen is also outdated and should be renewed.

PR #309, opened 2015-03-19
fix: Navigation drawer close issue

PareshMayani:
Issue:
When the navigation drawer is open and if you press back key, It closes the application (if home screen) or goes back to the previous screen.

Fix: 
Navigation drawer should be closed on back key press. I have included close navigation drawer code inside the onBackPressed() of GdgNavDrawerActivity, as it's a base activity.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
@@ -230,6 +230,25 @@ public abstract class GdgNavDrawerActivity extends GdgActivity {
         maybeUpdateChapterImage();
     }
 
+    @Override
+    public void onBackPressed() {
+        if (isNavDrawerOpen()) {
+            closeNavDrawer();
+        } else {
+            super.onBackPressed();
+        }
+    }
+
+    protected boolean isNavDrawerOpen() {
+        return mDrawerLayout != null && mDrawerLayout.isDrawerOpen(Gravity.START);
+    }
+
+    protected void closeNavDrawer() {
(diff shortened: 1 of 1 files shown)


tasomaniac:
Can you explain what was broken? I believe DrawerLayout should take care of this.

PareshMayani:
@tasomaniac I have updated issue details, please check.

tasomaniac:
I see. I confirmed that we have the issue. But in `DrawerLayout`, they have the following code to take care of back button automatically. I wonder why it doesn't work. 


    @Override
    public boolean onKeyDown(int keyCode, KeyEvent event) {
        if (keyCode == KeyEvent.KEYCODE_BACK && hasVisibleDrawer()) {
            KeyEventCompat.startTracking(event);
            return true;
        }
        return super.onKeyDown(keyCode, event);
    }

    @Override
    public boolean onKeyUp(int keyCode, KeyEvent event) {
        if (keyCode == KeyEvent.KEYCODE_BACK) {
            final View visibleDrawer = findVisibleDrawer();
            if (visibleDrawer != null && getDrawerLockMode(visibleDrawer) == LOCK_MODE_UNLOCKED) {
                closeDrawers();
            }
            return visibleDrawer != null;
        }
        return super.onKeyUp(keyCode, event);
    }

PareshMayani:
Even I was wondered about it but I double checked it with the IO app. They also have written code inside their base activity.

Check: https://github.com/google/iosched/blob/cf1f30b4c752f275518384a9b71404ee501fc473/android/src/main/java/com/google/samples/apps/iosched/ui/BaseActivity.java

tasomaniac:
Wow. Thank you.

Issue #270, opened 2015-03-02
Redesign EventDetail page.


tasomaniac:
Improve the use of card views in EventDetail.

PareshMayani:
I am working on this issue and have made improvements.

Can you please check the these snaps and let me know if any changes are required?

1st improvement: Where I have done layout improvements.
![screenshot_2015-03-20-00-37-29](https://cloud.githubusercontent.com/assets/888080/6746436/2e578d62-ceee-11e4-944d-08d4e8c98c25.png)

2nd improvement: Where I have removed that resource background from description layout and have used CardView, to make it looks neat and clean. It would also make all the description layouts consistent across the app.

![screenshot_2015-03-20-11-55-43](https://cloud.githubusercontent.com/assets/888080/6746994/5c651b48-cef8-11e4-8996-6ea8877be2e7.png)

Please let me know if I should commit these improvements and pull the request :)

tasomaniac:
Can you align the text inside the card and the texts in the header. Thank you. Great work.

PareshMayani:
Done.

![screenshot_2015-03-21-02-44-30](https://cloud.githubusercontent.com/assets/888080/6760991/6102eed8-cf74-11e4-8a6b-29d48bb5e77a.png)

friedger:
Closing this as it is not very actionable.

tasomaniac:
This can be a label for other Event Detail page related issues. Like `Event Detail`
We have so much thing to do to improve that page.

friedger:
nice, for the label. I would use component instead of epic.

tasomaniac:
makes sense. changing it now.

PR #312, opened 2015-03-20
Redesign event details layout

PareshMayani:
Resolution for the https://github.com/gdg-x/frisbee/issues/270

screenshot_2015-03-21-02-44-30


PareshMayani:
Thanks @friedger for accepting and merging.

tasomaniac:
Me being a ninja. :) I think the texts should be aligned. Header and the card is aligned here. CardView is tricky.

PareshMayani:
Texts are aligned on the left hand side. Header and card is not aligned, there is a difference of margin and padding. 
 
Please share more about the point if I haven't got it exactly.

tasomaniac:
I think it is not important. There is a slight padding difference I think. At most 2dp. Nevermind :)

friedger:
It matters. But this could be part of another PR.
Am 24.03.2015 08:51 schrieb "Said Tahsin Dane" <notifications@github.com>:

PR #313, opened 2015-03-24
refactor: card views in chapter info layout

PareshMayani:
1. Replaced people_box, resources_box and story_box with CardView.
2. Fix: clicking on the home chapter image results in clicks on the underlying activity


friedger:
would be nice to have a screenshot in the PR.

app/src/main/res/layout/navdrawer.xml
```diff
@@ -22,6 +22,7 @@
   android:layout_height="match_parent"
   android:orientation="vertical"
   android:layout_gravity="start"
+  android:clickable="true"
```
friedger: :+1:

PareshMayani:
## Before:
screenshot_2015-03-21-13-18-30

## After:
screenshot_2015-03-24-22-59-20

## Changes:
As we have used CardView everywhere, even chapter info portrait layout doesn't have those backgrounds. I think someone has forgot to remove those background from landscape layout, so removed those layouts and made minor re-factoring in both the layouts.

BTW, even I believe in 1 image = 1000 words rule :) but somehow I missed to post image!

tasomaniac:
All the old card view backgrounds removed and replaced by CardView rapidly. Landscape layout left out accidentially. Thank you for fixing that.

app/src/main/res/layout/navdrawer.xml
tasomaniac: What does this do? Does it fix the problem where we were able to click things underneath?

tasomaniac:
:+1:
As usualy I would say texts should be aligned. But that is for another PR.

PR #314, opened 2015-03-30
refactor: Removed redundancy and have included empty view layout

PareshMayani:
Empty view layout was already created but not used in layouts, removed redundancy from some of the layouts and have included empty view layout.


tasomaniac:
:+1:

PR #316, opened 2015-04-02
Library updates

tasomaniac:


tasomaniac:
BTW, analytics and gcm were part of the gms core. But they are now separate dependencies. Thats why I added them.

Splaktar:
Good to know. Guess that I will need to do this same kind of thing for my apps. The modularity will definitely help with the memory usage and dex counts.

tasomaniac:
They are getting better and better.

PR #319, opened 2015-04-03
chore: String constants are not actually constants.

### LINKED ISSUE #189: Define constants to use for putting and retrieving arguments
In my company's A-Team, we always define constants for putting and retrieving values from `getArguments()`.

I've noticed it's not practiced here: https://github.com/gdg-x/frisbee/pull/188#discussion_r24403811

I'd like to fix this. What do you think guys if (e.g. in [this file](https://github.com/HenriqueRocha/frisbee/blob/feature/186_pull_to_refresh/app/src/main/java/org/gdg/frisbee/android/fragment/GdlListFragment.java)) instead of:
```
arguments.putString("category", cat);
getArguments().getString("category")
```
would be:
```
private static final String ARGS_KEY_CATEGORY = "category"
arguments.putString(ARGS_KEY_CATEGORY, cat);
getArguments().getString(ARGS_KEY_CATEGORY)
```

tasomaniac:
I hope I didn't miss anything. I tried not to :)

Also fixes #189

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/ArrowActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/ArrowActivity.java
@@ -66,6 +66,7 @@ import butterknife.InjectView;
 
 public class ArrowActivity extends GdgNavDrawerActivity {
 
+    private static final Charset CHARSET = Charset.forName("US-ASCII");
     public static final String ID_SEPARATOR_FOR_SPLIT = "\\|";
     public static final String ID_SPLIT_CHAR = "|";
     private static final int REQUEST_LEADERBOARD = 1;
@@ -414,9 +415,9 @@ public class ArrowActivity extends GdgNavDrawerActivity {
                 String msg = getEncryptedMessage();
                 NdefRecord mimeRecord = new NdefRecord(
                         NdefRecord.TNF_MIME_MEDIA,
-                        Const.ARROW_MIME.getBytes(Charset.forName("US-ASCII")),
+                        Const.ARROW_MIME.getBytes(CHARSET),
                         new byte[0],
-                        msg.getBytes(Charset.forName("US-ASCII")));
+                        msg.getBytes(CHARSET));
                 return new NdefMessage(new NdefRecord[]{mimeRecord});
             } catch (Exception e) {
                 e.printStackTrace();
--- a/app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
@@ -70,6 +70,8 @@ public class FirstStartActivity extends ActionBarActivity implements
         FirstStartStep2Fragment.Step2Listener, 
         FirstStartStep3Fragment.Step3Listener {
 
+    public static final final String ACTION_FIRST_START = "finish_first_start";
+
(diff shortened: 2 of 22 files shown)


app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
```diff
@@ -70,6 +70,8 @@
         FirstStartStep2Fragment.Step2Listener, 
         FirstStartStep3Fragment.Step3Listener {
 
+    public static final final String ACTION_FIRST_START = "finish_first_start";
```
friedger: Travis spotted this "final final"

tasomaniac: Sorry about that. I doesn't compile locally either. Fixing now.

tasomaniac:
Building now.

app/src/main/java/org/gdg/frisbee/android/activity/ArrowActivity.java
```diff
@@ -66,6 +66,7 @@
 
 public class ArrowActivity extends GdgNavDrawerActivity {
 
+    private static final Charset CHARSET = Charset.forName("US-ASCII");
```
Splaktar: Using a more meaningful name like CHARSET_US_ASCII would be more clear. CHARSET is pretty generic for a final like this.

app/src/main/java/org/gdg/frisbee/android/fragment/FirstStartStep1Fragment.java
```diff
@@ -50,6 +51,7 @@
 
 public class FirstStartStep1Fragment extends Fragment {
 
+    private static final String ARG_SELECTED_CHAPTER = "selected_chapter";
```
Splaktar: Why is this redefined here? Can this definition in MainActivity be made public and reused here?

Splaktar:
I love the idea of using more constants and getting rid of repeated use of Strings. But would it make sense to move more of these constant definitions into Const.java?

Splaktar:
There also appear to be a handful of public constants in PrefUtils.java as well. The other *Utils classes don't seem to contain any public constants.

app/src/main/java/org/gdg/frisbee/android/activity/ArrowActivity.java
tasomaniac: I've missed in the huge refactoring. You are absolutely right.

app/src/main/java/org/gdg/frisbee/android/fragment/FirstStartStep1Fragment.java
tasomaniac: I thought about that and decided to define it twice because they are not actually related. They are for saving the state of the Activity and Fragment. And usually the constants for Intent extras and saving the instance state are defined in the respected files.

tasomaniac:
Thank you for your review @Splaktar Appreciated. 
PrefUtils has its own constants and actually it makes sense because they are not being used elsewhere. But actually there are some constants in the `Const` file related to `PrefUtils`. So it is kinda a mess and hard to get it right. 

We can decide and have all the constants in `Const` . What do the others think about this?

mauimauer:
Move it all to Const. So we define constants meaning the same thing only once (and have only one place to check).

PR #324, opened 2015-04-09
Gradle config updates

tasomaniac:
incremental is removed because it is dangerous. :P
textReport and textOutput to stdout is added. It will give lint errors in Travis.


erickmendonca:
:+1:

PR #325, opened 2015-04-09
update: proguard configuration

tasomaniac:
- Crashlytics is added.
- Proper and updated configs for okhttp, okio and butterknife
- ACRA removed


tasomaniac:
I will also merge this if no one says not to :)

app/proguard-rules.txt
```diff
@@ -68,49 +79,17 @@
...
-# keep this otherwise it is removed by ProGuard
--keep public class org.acra.ErrorReporter
-{
-    public void handleSilentException(java.lang.Throwable);
+-keepclasseswithmembernames class * {
+    @butterknife.* <methods>;
 }
 
-# GMS
+#GMS
```
Splaktar: Not sure why this was needed. There seems to be inconsistent spacing after `#` throughout the file. I prefer the space as do the snippets here (https://github.com/krschultz/android-proguard-snippets/tree/master/libraries).

tasomaniac: Thank you for the review. I will update it. Not sure why this happened. :)

tasomaniac: Do you think that we should use separate files for every library project we have? Where should we keep those files?

tasomaniac:
Just updated the PR. Can you take a look. ty.

PR #326, opened 2015-04-09
Sudo set to false to enable container-based infrastructure in Travis.

tasomaniac:
Read more here
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: android
 
+sudo: false
+
 android:
   components:
     - android-22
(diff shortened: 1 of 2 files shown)


tasomaniac:
It says faster builds but it fails because of a memory issue with an exit code of 137. 

I've been playing with this with my own toy project and making `threadCount 1` solved my problem. Since frisbee is much bigger project, I don't think we can solve it. 

Just closing this.

PR #328, opened 2015-04-22
fix: ioextended tag was wrong.

tasomaniac:
IO Extended section is not working and will not work when it is enabled. We should update the app before May 1st. 

perPage was not working and returning default 25 events. it should be small letters perpage. And appearantly -1 returns all the results

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/api/GdgXHub.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/GdgXHub.java
@@ -29,7 +29,7 @@ public interface GdgXHub {
     void getEventDetail(@Path("id") String eventId,
                         Callback<EventFullDetails> callback);
 
-    @GET("/events/tag/{tag}/upcoming?perPage=1000")
+    @GET("/events/tag/{tag}/upcoming?perpage=-1")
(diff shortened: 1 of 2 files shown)


tasomaniac:
We set Release 2.1 to May 5th. 

Should we release a bug fix release 2.0.4 before that?

friedger:
Let's release 2.0.4 first
Am 22.04.2015 16:57 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
Which issues should be included in 2.0.4?
Your new OrganizerFragment and my support 22.1 upgrade? Or should they wait for 2.1?

friedger:
Feel free to decide what is feasible..
Am 22.04.2015 18:35 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
Hey @friedger 

Adriana told that "I/O Extended" is the official tag. We cannot use multiple tags in the app. So what should be the real tag? I am not sure. I think "I/O Extended" is converted to "i-oextended"

There are 2 tag slugs in the GDG[X] Hub: ioextended and i-oextended

PR #330, opened 2015-04-28
fix: IO Extended tag is updated again

tasomaniac:
There are more events with "i-oextended" tag.


friedger:
@tasomaniac FYI You can look up all tags at https://developers.google.com/tags/gettags

tasomaniac:
I didn't know. Thank you. :) There are lots of hidden services in
developers.google.com :)

PR #331, opened 2015-04-28
Feature/Add leads fragment

friedger:
This PR adds a fragment that contains messages only for organizers.
Currently, the messages are hard coded.

device-2015-04-28-143919
device-2015-04-28-143937
device-2015-04-28-143949


app/src/main/java/org/gdg/frisbee/android/activity/GdgActivity.java
```diff
@@ -172,17 +172,17 @@ protected void onActivityResult(int requestCode, int responseCode, Intent intent
                         // If the error resolution was successful we should continue
                         // processing errors.
                         mSignInProgress = STATE_SIGN_IN;
+
```
tasomaniac: What does this move do?

app/src/main/java/org/gdg/frisbee/android/activity/MainActivity.java
```diff
@@ -379,7 +380,11 @@ public int getItemPosition(Object object) {
 
         @Override
         public int getCount() {
-            return 3;
+            if (App.getInstance().isOrganizer()) {
+                return 4;
```
tasomaniac: While we are doing this, I think we can also get rid of this hard coded numbers. We can have an array to store the name of the tabs, and count would be the size of that array.

app/src/main/java/org/gdg/frisbee/android/fragment/LeadFragment.java
```diff
@@ -0,0 +1,94 @@
...
+
+            ViewHolder viewHolder = (ViewHolder) convertView.getTag();
+            viewHolder.title.setText(getItem(position).getTitle());
+            viewHolder.details.setText(getItem(position).getDetails());
+
+            return convertView;
+        }
+
+        private static class ViewHolder {
+            TextView title;
```
tasomaniac: Can you use Butterknife here to make it more consistent w/ other Adapters and make it more beautiful?

app/src/main/res/values/strings.xml
```diff
@@ -152,4 +152,11 @@
     <string name="event_more">More</string>
     <string name="help">Help</string>
     <string name="feedback">Feedback</string>
+    <string name="for_leads">For Leads</string>
+    <string name="leads_welcome">Congrats for being a GDG organizer!</string>
+    <string name="leads_resources">You should check the GDG resource at https://drive.google.com/drive/#folders/0B55wxScz_BJtWW9aUnk2LUlNdEk</string>
```
tasomaniac: Since user cannot copy the link here and cannot memorize it, I don't see any usefulness of the link in the detail text. 
I think we can make it like "click here to access GDG resources"

app/src/main/java/org/gdg/frisbee/android/lead/LeadMessage.java
```diff
@@ -0,0 +1,48 @@
...
+    private String mLinkUrl;
+
+    private LeadMessage(Type type, final String title, final String details) {
+        mTitle = title;
+        mDetails = details;
+        mType = type;
+    }
+
+    private LeadMessage(final Type type, final String title, String details, final String linkUrl) {
+        this(type, title, details);
```
tasomaniac: Instead of this constructor using the other, I would make the other to use this constructor with a null linkUrl. What do you think?

app/src/main/java/org/gdg/frisbee/android/fragment/LeadFragment.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void onActivityCreated(Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+        mAdapter = new LeadAnouncementAdapter(getActivity());
+        setListAdapter(mAdapter);
+        mAdapter.add(LeadMessage.newMessage(getString(R.string.leads_welcome_title),
+                getString(R.string.leads_welcome)));
+        mAdapter.add(LeadMessage.newResource(getString(R.string.leads_resources_title),
+                getString(R.string.leads_resources),
+                "https://drive.google.com/drive/#folders/0B55wxScz_BJtWW9aUnk2LUlNdEk"));
```
tasomaniac: This URL is also defined in Const.java file. I think this and the below one should only be defined in Const.java.

app/src/main/java/org/gdg/frisbee/android/fragment/LeadFragment.java
```diff
@@ -0,0 +1,94 @@
...
+import org.gdg.frisbee.android.Const;
+import org.gdg.frisbee.android.R;
+import org.gdg.frisbee.android.lead.LeadMessage;
+
+public class LeadFragment extends ListFragment {
+    private LeadAnouncementAdapter mAdapter;
+
+    @Override
+    public View onCreateView(final LayoutInflater inflater, @Nullable final ViewGroup container, @Nullable final Bundle savedInstanceState) {
+        View root = inflater.inflate(R.layout.fragment_for_leads, null);
```
tasomaniac:
```
View root = inflater.inflate(R.layout.fragment_for_leads, container, false);
```
should be used instead

app/src/main/java/org/gdg/frisbee/android/fragment/LeadFragment.java
friedger: These are only temporary hard coded messages. They should come from the server in the future. Could be extracted into a AnnouncementFactory

app/src/main/java/org/gdg/frisbee/android/activity/GdgActivity.java
friedger: This fixes the infinite loop, or at least it does not try to connect when the user presses "Cancel"

tasomaniac: oh. makes sense.

app/src/main/java/org/gdg/frisbee/android/fragment/LeadFragment.java
tasomaniac: I know. But for now (if we are going to merge it) we shouldn't have duplicate String constants.

tasomaniac:
Thank you. I think this will be really really useful for organizers.
The build failure is about ButterKnife. It says that ViewHolder should be a public class.

tasomaniac:
I think when you merged mine, it created conflicts. If you can resolve the conflicts, I can merge this.

tasomaniac:
@friedger 1 question that I've just realized. Suppose that the Main screen is open and I did a login somehow (e.g via settings screen) we should update the tabs on the main screen. Plus, we may display the LEAD Welcome tab for the first login as a welcome. 

I think we are not doing it now but I am not sure. I cannot login to the app with my credentials. :)

Issue #234, opened 2015-02-26
Add visual highlight to the material nav drawer


friedger:
There should be an image of the home gdg or just the logo of the app if no home gdg was selected.
The users account name should be visible as well.

tasomaniac:
As you said before, I also think that there is no need for a user switcher. 
But having the account name and using that account name area for login would be great. Because right now, if a user skips at setup, they have to go to the settings.

HenriqueRocha:
I worked on this today and will work more later today. Hopefully I'll be able to submit a PR for this.

friedger:
Is the image not enough? What are your plans?


  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-03-01 18:13 GMT+01:00 Henrique Rocha <notifications@github.com>:

HenriqueRocha:
@friedger The avatar of the user should be displayed also just like any other app that uses the drawer with login.
Another thing that I already did but still needs some touches is placing the status bar on top of the drawer.

friedger:
Good, go ahead, but I think the user's avatar is not a must for 2.0


  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-03-01 22:06 GMT+01:00 Henrique Rocha <notifications@github.com>:

HenriqueRocha:
Okay, I won't add it now then. I'll finish this and move to work on the CardViews which should be removed almost everywhere. I already did some work on that also.

tasomaniac:
I did Event Fragment and thinking to finish evet detail and info today

friedger:
closed by #333

PR #333, opened 2015-04-28
Feature/Add user picture to nav drawer

friedger:
This PR adds the user picture to the nav drawer when logged in.
It is a basic implementation and does not cover edge cases.
When not logged in a grey avatar is shown.

device-2015-04-29-102752

device-2015-04-28-234557


friedger:
This is the last part for #234

stefanhoth:
That looks great!

Could you add a screenshot for the logged out state?

app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
```diff
@@ -249,10 +259,22 @@ protected void closeNavDrawer() {
         }
     }
 
+    protected void updateUserPicture() {
+        com.google.android.gms.plus.model.people.Person user = com.google.android.gms.plus.Plus.PeopleApi.getCurrentPerson(getGoogleApiClient());
```
stefanhoth: Why the full package name?

tasomaniac:
It is a little bit detail but the icons of the list items and the user profile logo are left aligned in Google's implementation.

And we can add users name and email just like in the Google's implementation without a Account switcher Spinner. 

And since we have lots of items in the Nav Drawer, I think we should split them right? That can be another issue though.

app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
friedger: we are also using the com.google..api... classes

friedger:
user name and email might be difficult to read depending on the chapter image. For the profile we have always a white border.

friedger:
Screenshots updated

tasomaniac:
@friedger then it is a good reason to use new fancy ColorUtils class from support library to find contrasting text color :)
https://chris.banes.me/2015/04/22/support-libraries-v22-1-0/

friedger:
@tasomaniac I think that is too much for this release. Finding the right color of the area below the text, etc...

friedger:
added #338

tasomaniac:
I was just flying on the clouds. (A Turkish idiom for suggesting not urgent and hard things) :smile: 

BTW, I think we can remove not-logged-in avatar completely for now and merge this. The avatar will be a great improvement but it looks ugly when it is not rounded.

friedger:
Screenshot updated.

tasomaniac:
I think it is really nice. ty

PR #334, opened 2015-04-28
Support lib update and Material Dialogs

### LINKED ISSUE #327: Upgrade to new support library
And change all Dialogs to use `AppCompatDialog`

tasomaniac:
- Support library is updated to 22.1.1
- Uses Doorbell's gradle aar library
- Doorbell Dialog recreated with these advantages:
    - Material Dialog FTW. The same beautiful look for all devices.
    - It uses DialogFragment which much more reliable.
    - With FloatLabelLayout from Chris Banes, it is more understandable with labels.
    - Strings are extracted for translation purposes.
    - `AutoCompleteTextView` is used for Email field to support suggestions of email addresses. Finally!

This is how it looks on KitKat emulator: (Should I make the title the accent color?)
screen shot 2015-04-29 at 00 02 02
screen shot 2015-04-29 at 00 11 05

Closes #327


tasomaniac:
The Travis Error is about translations of the new strings. We should update crowdin and download translations again. The credentials are on my other computer. I can do it tomorrow. 

Or we can use `warning 'MissingTranslation'` to ignore it but I don't like to do that. It is the same problem with #331

app/src/main/java/org/gdg/frisbee/android/activity/FirstStartActivity.java
```diff
@@ -65,7 +65,7 @@
 import retrofit.client.Response;
 import timber.log.Timber;
 
-public class FirstStartActivity extends ActionBarActivity implements 
+public class FirstStartActivity extends AppCompatActivity implements
```
friedger: :+1:

friedger:
Let's merge both

Issue #335, opened 2015-04-28
Drawer behind status bar. 


tasomaniac:
Nav Drawer should extend to the top and go behind the status bar. 

Like this:
![](http://1.bp.blogspot.com/-Ychs6FLcjGA/VE6og8WgM9I/AAAAAAAAA78/uSFiUm5V8kM/s400/drawerlayering.gif)

PR #337, opened 2015-04-28
refactor: themes

tasomaniac:
There is no need for `Theme.GDG.NoActionBar` while all the Activities using the same theme. So we have now only `Theme.GDG`

`values-v21/gdg-themes.xml` is kept becuase it will be used for issue #335


friedger:
@tasomaniac oh, some merge conflicts, can you fix that?

tasomaniac:
Yep :) I did it just now. :)

tasomaniac:
shamelessly merged my PR because I need this before I work on #335

PR #341, opened 2015-04-29
Libraries are updated

tasomaniac:


friedger:
might be nice to see in the PR description what the new versions bring (if any)

tasomaniac:
Right. The changelog of Play Services is not actually directly related to us. But there are of course bug fixes on every release. So I think it is nice to be up to date.

PR #346, opened 2015-04-30
Fix/Model cache for ArrayList

friedger:
This PR removes use of ArrayList and instead uses subclasses of ArrayList.

The ModelCache could not restore the component class of lists because the generic type information was in https://github.com/gdg-x/frisbee/blob/1cdbaa6d0c7575ad4eb79ad4762ece6faa8f46ff/app/src/main/java/org/gdg/frisbee/android/utils/Utils.java#L71 was erased.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/cache/ModelCache.java
+++ b/app/src/main/java/org/gdg/frisbee/android/cache/ModelCache.java
@@ -579,30 +579,6 @@ public class ModelCache {
             } catch (ClassNotFoundException e) {
                 e.printStackTrace();
             }
-        } else if (className.startsWith("java.util.HashMap<")) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/gdg/frisbee/android/cache/ModelCache.java
```diff
@@ -605,8 +566,6 @@ private Object readValueFromDisk(InputStream is) throws IOException {
             if (className.contains("google")) {
```
tasomaniac: What about this? :) Did you find what is this for?

tasomaniac:
I think it looks good.

tasomaniac:
Failed because of an unused import :)

friedger:
All green :-)

PR #347, opened 2015-05-03
Fix typo in README. Add reference to debug signing config.

Splaktar:


tasomaniac:
Oh thank you for the signingconfig fix.

PR #348, opened 2015-05-04
Drawer behind statusbar

tasomaniac:
Here is the screenshot of the new implementation. 

I've looked at the sizes of the nav drawer in the Google IO app and adjusted the image sizes. 
- The profile image is `40dp/64dp`
- The chapter image height `140dp/164/dp`
- The nav drawer width `260dp/304dp`

respectively on phones and devices with min `360dp` width.

Additionally I've adjusted the themes into 2 categories just like Google IO app. 
We have now 
`Theme.GDG` and `Theme.GDG.WithNavDrawer`

screen shot 2015-05-04 at 16 24 56

It closes issue #335


app/src/main/res/values/gdg_themes.xml
```diff
@@ -42,11 +44,32 @@
...
+  <style name="Theme.GDG.Main" parent="Theme.GDG.WithNavDrawer">
     <item name="actionBarInsetStart">@dimen/keyline_2_minus_16dp</item>
   </style>
 
-  <style name="Theme.GDG.Special.Wtm" parent="Theme.GDG">
+  <!-- Special Event Series-->
+  <style name="Theme.GDG.Special" parent="Theme.GDG.WithNavDrawer"/>
+
+  <!-- Experts Directory -->
+  <style name="Theme.GDG.GDE" parent="Theme.GDG.WithNavDrawer"/>
```
friedger: Do we need these? I would prefer to add them when they are needed to reduce the code that needs to be maintained.

app/src/main/AndroidManifest.xml
```diff
@@ -93,14 +98,15 @@
...
+      android:screenOrientation="portrait"
+      android:theme="@style/Theme.GDG.Arrow"/>
     <activity
       android:name=".activity.FirstStartActivity"/>
     <activity
       android:name=".activity.YoutubeActivity"
       android:configChanges="keyboardHidden|orientation|screenSize"
       android:screenOrientation="sensor"
-      android:theme="@style/GDG.SeeTrough"/>
+      android:theme="@style/Theme.GDG.SeeTrough"/>
```
friedger: :+1: for proper class hierarchy

app/src/main/res/values/gdg_themes.xml
tasomaniac: I thought about it really much and I ended up doing it because Google IO app did it like this. They have empty styles for different activities. I also did that to get your opinion.

app/src/main/AndroidManifest.xml
tasomaniac: Thank you :)

app/src/main/res/values/gdg_themes.xml
friedger: Currently, I don't see any advantages in having empty styles.

tasomaniac: One advantage is that we have theme names in the Manifest that we never change. Otherwise they all become `Theme.GDG` and `Theme.GDG.WithDrawer`
The other might be that when we look at the themes.xml, we are seeing all the styles at once.

PR #349, opened 2015-05-04
Refactor: DEVELOPER_MODE was not used.

tasomaniac:


friedger:
probably, we can get rid of Alpha flavor as well.

Issue #68, opened 2013-08-28
Widget does not have a button to enter the GDG app


friedger:
There should be an area to click to open the GDG app

tasomaniac:
I think the best way to do this to open the event detail page in the app rather than directly openning the Google+ app. Easy and clean. Would you prefer doing that? @friedger 

Otherwise we need to change the design of the widget. I am sure that it is not used much.

friedger:
Sounds good to me, I'll do it.


  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-05-04 21:38 GMT+02:00 Said Tahsin Dane <notifications@github.com>:

PR #350, opened 2015-05-05
68/Add a default click event and always open the GDG app from app widget

friedger:
With the event details page we always want to open the GDG app.


tasomaniac:
Looks good :) This also closes #68 
But as far as I see here, it makes a network request on every widget update. Is it true?

friedger:
events are not cached, yes. Also not on the event details activitiy


  [image: --]
Friedger Müffke
<http://about.me/friedger>


2015-05-05 12:42 GMT+02:00 Said Tahsin Dane <notifications@github.com>:

tasomaniac:
I am not talking about the event detail.

Event list is cached though. We should get the cached event list. I think we should merge this and deal with it later. Because we now have an issue about removing the cache and use HTTP cache. If we use HTTP cache my comment here will be resolved automatically.

PR #352, opened 2015-05-08
update: support and play services libraries are now using a variable.

tasomaniac:
plus: really small library updates.


friedger:
:+1:

PR #356, opened 2015-05-08
crash fixes

tasomaniac:
new leak canary integrated.
Using BaseFragment with leak canary integration.
Has run monkey and found couple of easy crashes.
Crashlytics is already disabled for debug builds, it is not completely disabled to improve build speed.


app/src/main/java/org/gdg/frisbee/android/event/EventOverviewFragment.java
```diff
@@ -191,6 +191,10 @@ public void failure(RetrofitError error) {
     }
 
     private void updateGroupDetails(Chapter group) {
+        if (getActivity() == null) {
```
friedger: While this solves the issue I think the problems is somewhere else.

app/src/main/java/org/gdg/frisbee/android/fragment/InfoFragment.java
```diff
@@ -220,13 +219,19 @@ public void onClick(View view) {
                 }
             });
             registerForContextMenu(v);
-            mOrganizerBox.addView(v);
+            if (mOrganizerBox != null) {
```
friedger: I would expect the contract for this method that the UI is already setup.

app/src/main/java/org/gdg/frisbee/android/fragment/InfoFragment.java
```diff
@@ -220,13 +219,19 @@ public void onClick(View view) {
...
+                mOrganizerBox.addView(v);
+            }
         }
     }
 
     private void updateChapterUIFrom(final Person person) {
-        mTagline.setText(person.getTagline());
-        mAbout.setText(getAboutText(person));
+        if (mTagline != null) {
```
friedger: Same here..

app/src/main/java/org/gdg/frisbee/android/event/EventOverviewFragment.java
tasomaniac: The problem is with Fragments in general. Since there is an async operation going on, when user press back before the operation completes, Fragment looses its tie with the Activity. Other option to solve this is to cancel all async operations in onDestroy

app/src/main/java/org/gdg/frisbee/android/fragment/InfoFragment.java
tasomaniac: This actually happens after onDestroyView is called.

PR #357, opened 2015-05-14
Login dialog

### LINKED ISSUE #222: Open Login page with a description for things that require login.
Instead of giving an error when the button is clicked, 

open the Login page with an information about the clicked button, the feature and why login is necessary.

tasomaniac:
Signin Needed dialog support. 
- A dialog is shown to users to take them to sign-in when needed. (It is now when Arrow and Achivements clicked)
- After a succesful login, they are also taken to the item clicked.

Fixes #222 

**Side note:** Cannot login with debug builds. I've tested it with release builds and was working. 
But after openning Arrow page, it couldn't detect that I am an organizer. It was because of a 400 error from googleapis.com. Production work fine.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/activity/GdgNavDrawerActivity.java
@@ -16,12 +16,15 @@
 
 package org.gdg.frisbee.android.activity;
 
+import android.content.DialogInterface;
 import android.content.Intent;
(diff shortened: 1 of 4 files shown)


tasomaniac:
Sorry all. I forgot to add screenshots. Here they are:

screen shot 2015-05-13 at 11 54 02
screen shot 2015-05-13 at 11 53 43

tasomaniac:
Merging this since it is a small PR and it is so much better than the previous implementation :)

PR #381, opened 2015-07-21
Makes the debug version installable separately.

tasomaniac:
Debug suffixes added to applicationId, versionName and application name.
LeakCanary Activity label is now "GDG Leaks"

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,14 +21,14 @@
   <uses-permission android:name="android.permission.GET_TASKS"/>
 
   <permission
-    android:name="org.gdg.frisbee.android.permission.MAPS_RECEIVE"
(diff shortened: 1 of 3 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -21,14 +21,14 @@
   <uses-permission android:name="android.permission.GET_TASKS"/>
 
   <permission
-    android:name="org.gdg.frisbee.android.permission.MAPS_RECEIVE"
+    android:name="${applicationId}.permission.MAPS_RECEIVE"
```
PareshMayani: that's a good fix :)

tasomaniac: Thank you but Travis is failing because it cannot find the commit. I have to fix that.

PR #385, opened 2015-07-22
Retry button is added and duration is made infinite.

tasomaniac:
Here is how it looks. It is definitely out of guidelines. By default the Action color is the accent color. When we color the background, there is only little possibility that accent color will match with red, green, blue etc. So I made it all white for all cases. 

We definitely need a RETRY button there. We can do it like below. Or we can use the standard Snackbar. 

shamulyz28etasomaniac07222015134655

shamulyz28etasomaniac07222015135052

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/onboarding/FirstStartStep1Fragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/onboarding/FirstStartStep1Fragment.java
@@ -134,7 +134,13 @@ public class FirstStartStep1Fragment extends BaseFragment {
             public void failure(RetrofitError error) {
                 try {
                     Snackbar snackbar = Snackbar.make(getView(), R.string.fetch_chapters_failed,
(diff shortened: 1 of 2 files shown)


PareshMayani:
Retry button is indeed required for this screen.

Splaktar:
I prefer the standard snackbar. Making the retry text red or yellow may be appropriate in this case.

tasomaniac:
I realized that they have color functions now. https://developer.android.com/sdk/support_api_diff/22.2.1/changes.html

friedger:
LGTM

tasomaniac:
Thank you for the merge. 
@Splaktar the theming of the Snackbar is I think out of this PR's scope. What do you think we should do? It is an easy task actually.

Splaktar:
Yeah, it looks like we can use the default background for the snackbar and use Snackbar SetActionTextColor(ColorStateList) or Snackbar setActionTextColor(int) to set the `Retry` text color to red.

PR #386, opened 2015-07-22
Old PreferenceFragment implementation is removed.

tasomaniac:
It uses reflection and not needed anymore as we support API Level 14+
For a weird reason, there was a casting error occurred when static fragment is used. So I added it dynamically.

Look and feel is exactly the same.


friedger:
Less code is always good :-)

PR #390, opened 2015-07-24
Organiser Check is fixed

### LINKED ISSUE #360: Arrow tagging layout loses button, breaks send mode
After updating to the latest version, the arrow button is no longer there. This means that I can tap the screen and do a visual scan to tag others, but I am completely unable to allow others to tag me (scan mode).

I was able to get it back by reinstalling the app, but after I scanned someone else (Said) the button disappeared again. Logging out of G+ and back in does not fix it.

I was trying to get more people using the app and tagging each other now at the summit, but this is really breaking that :disappointed:

tasomaniac:
- It was caching for 30 days in SharedPrefs. Cache mechanism is completely removed. I can re-add it if someone wants. I don't think it is necessary. New organisers should see them immediately.
- When the currentId is null, which happens on app start, it was making the request. Since the request fails, it thoughts that the user is not organiser.

Fixes #360

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/Const.java
+++ b/app/src/main/java/org/gdg/frisbee/android/Const.java
@@ -61,9 +61,6 @@ public class Const {
 
     //Arrow
     public static final String QR_MSG_PREFIX = "gdgx://arrow?m=";
-    public static final String PREF_ORGANIZER_CHECK_TIME = "pref_organizer_check_time";
-    public static final String PREF_ORGANIZER_CHECK_ID = "pref_organizer_check_id";
-    public static final long ORGANIZER_CHECK_MAX_TIME = 2592000000L; // 30 days
     public static final String ARROW_MIME = "application/vnd.org.gdgx.frisbee.arrow";
     public static final String ARROW_LB = "CgkIh5yNxL8MEAIQBw";
     public static final String ARROW_K = "XXXX111122223333";
--- a/app/src/main/java/org/gdg/frisbee/android/Const.java
+++ b/app/src/main/java/org/gdg/frisbee/android/Const.java
@@ -61,6 +61,10 @@ public class Const {
 
     //Arrow
     public static final String QR_MSG_PREFIX = "gdgx://arrow?m=";
+    public static final String PREF_ORGANIZER_CHECK_TIME = "pref_organizer_check_time";
+    public static final String PREF_ORGANIZER_CHECK_ID = "pref_organizer_check_id";
+    public static final String PREF_ORGANIZER_STATE = "pref_organizer_state";
+    public static final long ORGANIZER_CHECK_MAX_TIME = 2592000000L; // 30 days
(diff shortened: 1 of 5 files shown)


friedger:
Most of the users are not organisers, so caching should do good to our users.
Maybe we can add an item in the settings, like "I am a organiser now! Please update me."

friedger:
Why did the request fail?

tasomaniac:
@Splaktar When the application is first opened, GoogleApiClient is not connected yet and without waiting it, it tries a login with a null gplusId. And it failes.

tasomaniac:
@friedger I will add the cache again. How long the cache duration should be? What do you think? 30 days is too much I think.

Splaktar:
I think that 7 days is plenty and should keep from draining battery or having other negative effects for users.

tasomaniac:
It is now done with 1 week of caching. 
And I also fixed an exception I got while checking the offline cache. 

Should we put a button to force update in Settings? I don't think it is that needed.

app/src/main/java/org/gdg/frisbee/android/Const.java
```diff
@@ -63,7 +65,8 @@ private Const() {
     public static final String QR_MSG_PREFIX = "gdgx://arrow?m=";
     public static final String PREF_ORGANIZER_CHECK_TIME = "pref_organizer_check_time";
     public static final String PREF_ORGANIZER_CHECK_ID = "pref_organizer_check_id";
-    public static final long ORGANIZER_CHECK_MAX_TIME = 2592000000L; // 30 days
+    public static final String PREF_ORGANIZER_STATE = "pref_organizer_state";
+    public static final long ORGANIZER_CHECK_MAX_TIME = DateUtils.WEEK_IN_MILLIS;
```
Splaktar: :+1:

Splaktar:
LGTM

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -67,13 +68,18 @@ public void failure(RetrofitError error) {
...
-                    mAdapter.addAll(events);
-                    setIsLoading(false);
-                    Snackbar snackbar = Snackbar.make(getView(), R.string.cached_content,
-                            Snackbar.LENGTH_SHORT);
-                    ColoredSnackBar.info(snackbar).show();
+                    if (checkValidCache(item)) {
+                        ArrayList<Event> events = (ArrayList<Event>) item;
+
+                        mAdapter.addAll(events);
+                        setIsLoading(false);
```
friedger: Should be outside the if clause, shouldn't it?

tasomaniac: You are right. I wish you didn't merge the PR. :smile: I will push the fix to develop branch. Please have a look.

PR #392, opened 2015-07-24
Refactor/391 tablayout

### LINKED ISSUE #111: Welcome the organizer of a GDG with a special message
When the user is one of the organizers of the user's home GDG set a flag and a message to congratulate for the great work.

In the future, the gdg manager will have an extra menu for additional actions.

### LINKED ISSUE #391: Use new TabLayout instead of SlidingTabLayout

tasomaniac:
`SlidingTabLayout` removed in favor of new `TabLayout` from support design library.

Fixes #391 and hopefully fixes #111


Splaktar:
Any difference in looks?

tasomaniac:
I forgot to put screenshots. The look is the same. 
There was a bug in the old SlidingTabLayout. It didn't update itself when the ViewPager adapter is updated. That's why the Leads section was not shown for the Lead Organizers.

PR #393, opened 2015-07-25
Remove browser api key

tasomaniac:
This work began with removing `browser_simple_api_access_key` and it included bunch of stuff. 

- `browser_simple_api_access_key` is removed. 
- `ip_simple_api_access_key` cannot be removed because G+ apis used in `NewsFragment` didn't work with `android_simple_api_access_key`
- Plus APIs used in the app are updated to be used from GMS. Only News Feed requests are done with the Google API Client. I wish we could remove it. 
- And finally got rid of the compiler warning by adding global exclude into configurations.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
@@ -33,14 +33,14 @@ import android.widget.LinearLayout;
 import android.widget.ScrollView;
 import android.widget.TextView;
 
-import com.google.api.services.plus.model.Person;
+import com.google.android.gms.plus.model.people.Person;
 
 import org.gdg.frisbee.android.Const;
 import org.gdg.frisbee.android.R;
-import org.gdg.frisbee.android.common.GdgNavDrawerActivity;
 import org.gdg.frisbee.android.app.App;
 import org.gdg.frisbee.android.cache.ModelCache;
 import org.gdg.frisbee.android.common.BaseFragment;
+import org.gdg.frisbee.android.common.GdgNavDrawerActivity;
 import org.gdg.frisbee.android.task.Builder;
 import org.gdg.frisbee.android.task.CommonAsyncTask;
 import org.gdg.frisbee.android.utils.Utils;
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/gdg/frisbee/android/gde/GdeAdapter.java
```diff
@@ -129,21 +113,10 @@ public Person doInBackground(ViewHolder... params) {
                         Person gde = null;
 
                         if (matcher.matches()) {
-                            String plusId = matcher.group(1);
-                            gde = (Person) App.getInstance().getModelCache().get(Const.CACHE_KEY_GDE + plusId, !Utils.isOnline(mContext));
```
tasomaniac: This part is also leaking memory. 

When the device is only, GDE profiles are being requested live from G+ api while user scrolls. And these actually keep the ListFragment reference. When the user scrolls through tabs quickly, it doesn't let the Fragment instances to be recycled. 

getPersonSync implementation is always using the cache when available. So it will be a little bit better.

tasomaniac: And it caches the image for 2 days.

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -259,11 +254,11 @@ private void navigateTo(Class<? extends GdgActivity> activityClass, Bundle addit
...
-        plusClient = new Plus.Builder(mTransport, mJsonFactory, null)
-                .setGoogleClientRequestInitializer(
-                        new CommonGoogleJsonClientRequestInitializer(BuildConfig.IP_SIMPLE_API_ACCESS_KEY))
-                .setApplicationName("GDG Frisbee")
-                .build();
+//        plusClient = new Plus.Builder(mTransport, mJsonFactory, null)
+//                .setGoogleClientRequestInitializer(
+//                        new CommonGoogleJsonClientRequestInitializer(BuildConfig.ANDROID_SIMPLE_API_ACCESS_KEY))
+//                .setApplicationName("GDG Frisbee")
+//                .build();
```
Splaktar: Why is left here if it is commented out?

tasomaniac: I've tried lots of things in this work. 
I tried to get rid of these mTransport, mJsonFactory etc. But at the end I give up. 
These are left by mistake. Fixing now.

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -120,14 +120,14 @@ public void onActivityCreated(Bundle savedInstanceState) {
 
         mInflater = LayoutInflater.from(getActivity());
 
+        final String chapterPlusId = getArguments().getString(Const.EXTRA_PLUS_ID);
```
friedger: This can also be from the URI. Better to have a callback to the activity.
But different PR :-)

tasomaniac: I think with your app indexing commit, this EXTRA_PLUS_ID will be here without a problem. :smile: 
Activity will give it to Fragment.

PR #394, opened 2015-07-25
Memory leak fix thanks to Leak Canary.

tasomaniac:
ListFragmentSwipeRefreshLayout is a static class now with a WeakReference of the RecyclerView in the Fragment.
Since it was used in the inner class, RecyclerView was being leaked.


Splaktar:
:+1:

friedger:
static inner classes are always good :+1:

tasomaniac:
This was the most frequent leak in LeakCanary. 
I have couple of other Leaks related to Google Play Services. Unfortunately GMS is leaking our Fragments. 
I think they should be added to Leak Canary repo.

PR #403, opened 2015-07-25
fix: Listview divider in leads tab

PareshMayani:
Listview divider was visible :)

screenshot_2015-07-25-23-28-58


Splaktar:
Have an after screenshot? I assume the one above is before?

tasomaniac:
What should we do with the the leads section title being two lines?

PareshMayani:
@tasomaniac will fix it up and send it in a PR!

tasomaniac:
@PareshMayani what is the suggested fix? It will be always 2 lines for small phones when we have 4 tabs.

PareshMayani:
@tasomaniac I just created a separate issue #414, let's discuss there!

PR #405, opened 2015-07-25
This should fix scroll problems with NewsFragment

### LINKED ISSUE #399: News: Can't scroll up, always tries to refresh
Running the latest debug build from develop, once I scroll down on the News screen, I can't scroll back up. It always tries to refresh the view instead of scrolling up.

tasomaniac:
I set the RecyclerView of the SwipeRefreshLayout on the wrong place. It is fixed now.
Fixes #399


Splaktar:
Nice! :+1:

tasomaniac:
Thank you Michael for reporting it.

PR #407, opened 2015-07-25
GoogleApiClient is in App class now. 

tasomaniac:
Google API Client is created in every Activity. Which is I think bad.
Since we are using it almost everywhere, I think it is better to define it once in Application class.

What do you think? If anyone have time, can you test it?

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
@@ -40,7 +40,7 @@ import java.util.List;
 import butterknife.ButterKnife;
 
 public abstract class GdgActivity extends TrackableActivity implements
-        GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener {
+        GoogleApiClient.OnConnectionFailedListener {
 
     private static final int STATE_DEFAULT = 0;
     private static final int STATE_SIGN_IN = 1;
@@ -97,10 +97,7 @@ public abstract class GdgActivity extends TrackableActivity implements
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
-        GoogleApiClient googleApiClient = getGoogleApiClient();
-        googleApiClient.registerConnectionCallbacks(this);
-        googleApiClient.registerConnectionFailedListener(this);
+        getGoogleApiClient().registerConnectionFailedListener(this);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
```diff
@@ -100,38 +89,24 @@ protected String getTrackedViewName() {
...
-                    .addApi(AppStateManager.API)
-                    .addScope(Plus.SCOPE_PLUS_LOGIN)
-                    .addScope(Plus.SCOPE_PLUS_PROFILE)
-                    .addScope(Games.SCOPE_GAMES)
-                    .build();
-        }
-
-        mAchievementActionHandler =
-                new AchievementActionHandler(getHandler(), mGoogleApiClient, this);
+        getGoogleApiClient().registerConnectionFailedListener(this);
```
friedger: This is leaking the activity, unregistration is necessary.

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -77,6 +78,13 @@
     private MenuItem drawerItemToNavigateAfterSignIn = null;
     private static final int GROUP_ID = 1;
 
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        getGoogleApiClient().registerConnectionCallbacks(this);
```
friedger: This is leaking the activity, unregistration is necessary.

friedger:
The documentation says to instantiate in the activity onCreate method.

friedger:
From http://stackoverflow.com/questions/16827839/access-google-plus-client-from-multiple-activities

>Each Activity should maintain its own instance of GoogleApiClient since it is a lightweight object used to access the shared state held by the Google Play services process. 

So, I'd keep it like it is now.

friedger:
Maybe we could use enableAutoMange https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.Builder#enableAutoManage(android.support.v4.app.FragmentActivity, int, com.google.android.gms.common.api.GoogleApiClient.OnConnectionFailedListener)

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
tasomaniac: Do you know what is the difference between doin it like this or using `addConnectionCallbacks` in the builder. When we do it in the builder we never unregister.

tasomaniac:
You are right. But I've been seeing lot of leakage because of GoogleApiClient already. The problem is that when you navigate through Nav Drawer we are opening every activity on top of each other. After 10 nav drawer clicks, we have to press back 10 times. This means 10 GoogleApiClient object being allocated. Instead of removing GoogleApiClient we can fix that issue.

PR #408, opened 2015-07-26
Use settings before launching tests, rename PrefUtils.setInitialSettings

friedger:
Tidy ups the tests


tasomaniac:
Thank you. Now we should run these tests on Travis with an emulator. We had memory problems in Travis already. Do you think it can be done?

PR #410, opened 2015-07-26
336/Add separator in NavigationView

friedger:
Adds a separator after Arrow menu entry.

Having more separator, e.g. for events is too much especially, if there is only one or two events.

Before | After
---|---
device-2015-07-26-105724|device-2015-07-26-105350

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
@@ -76,6 +76,7 @@ public abstract class GdgNavDrawerActivity extends GdgActivity {
 
     private MenuItem drawerItemToNavigateAfterSignIn = null;
     private static final int GROUP_ID = 1;
+    private static final int SETTINGS_GROUP_ID = 2;
 
     @Override
     public void setContentView(int layoutResId) {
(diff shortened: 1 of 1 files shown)


tasomaniac:
Great improvement? Why can't we add sections for others like special events and games.

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -134,11 +135,10 @@ private void setupDrawerContent(NavigationView navigationView) {
         menu.add(GROUP_ID, Const.DRAWER_ACHIEVEMENTS, Menu.NONE, R.string.achievements).setIcon(R.drawable.ic_drawer_achievements);
         menu.add(GROUP_ID, Const.DRAWER_ARROW, Menu.NONE, R.string.arrow).setIcon(R.drawable.ic_drawer_arrow);
 
-        menu.addSubMenu("");
-        menu.add(GROUP_ID, Const.DRAWER_SETTINGS, Menu.NONE, R.string.settings).setIcon(R.drawable.ic_drawer_settings);
-        menu.add(GROUP_ID, Const.DRAWER_HELP, Menu.NONE, R.string.help).setIcon(R.drawable.ic_drawer_help);
-        menu.add(GROUP_ID, Const.DRAWER_FEEDBACK, Menu.NONE, R.string.feedback).setIcon(R.drawable.ic_drawer_feedback);
-        menu.add(GROUP_ID, Const.DRAWER_ABOUT, Menu.NONE, R.string.about).setIcon(R.drawable.ic_drawer_about);
+        menu.add(SETTINGS_GROUP_ID, Const.DRAWER_SETTINGS, Menu.NONE, R.string.settings).setIcon(R.drawable.ic_drawer_settings);
```
PareshMayani: This change shows you deleted old lines and added new lines. Yes it's not a major thing but only group should be changed!

tasomaniac: I didn't understand Paresh. I think he only changed GROUP_ID part.

PareshMayani: @tasomaniac oops ....my mistake! yes it's correct that only GROUP_ID part is changed!

friedger:
We can have a separator for games, but I wouldn't add one for events. Events are the main part of the app.

tasomaniac:
@friedger makes sense. And when there is only 1 event series occurring in the world, it would be awkward to have section for only 1.

PareshMayani:
@tasomaniac that's what I was supposed to share and that's the reason I merged this PR. Another reason of not implementing section and creating a separate group for special events is, currently there is a bug in library that you can't keep the subgroup item selected, but only main group items!

tasomaniac:
I was investigating it and realized that we don't have selected states of the main items anyway. If you select Developer Experts for example, the menu items are all the same

PareshMayani:
@tasomaniac this is because I didn't declared any of the group checkable due to the bug I mentioned in 336 issue.

tasomaniac:
But we don't have sections. They should be checkable.
As far as I can see in the issue, submenus without title doesn't do any
harm.

friedger:
We don't event have submenus.


  [image: --]
Friedger Müffke
<http://about.me/friedger>


2015-07-26 17:01 GMT+02:00 Said Tahsin Dane <notifications@github.com>:

tasomaniac:
What about this? How does it look?

What do you all think? @friedger @HenriqueRocha @PareshMayani @Splaktar @stefanhoth 

screen shot 2015-07-27 at 00 48 11

PareshMayani:
@tasomaniac it looks good to me.

Splaktar:
:+1:

Issue #374, opened 2015-07-11
Should validate empty input


PareshMayani:
Issue is in Feedback dialog. If user clicks on a send button then it should validate inputs, before actually making a call.

tasomaniac:
Are we talking about just the email address validation? Or do we have
anything else?

jmsalcido:
as we are using Doorbell.io for the feedback I think we only should validate the email address, we already have a Utils.java that contains a isEmailAddress method that we could use before doing the call.

PareshMayani:
I am talking about empty input. We are making a call even though user leaves fields empty and click on send button. There should a proper message saying that "Please input message" or "Please input email address"!

PareshMayani:
@tasomaniac thanks, will fix it up!

friedger:
@PareshMayani What needs to be done for this one?

tasomaniac:
This is already fixed in the above referenced prs

PR #412, opened 2015-07-26
fix: validate inputs in feedback dialog

PareshMayani:
This fixes up https://github.com/gdg-x/frisbee/issues/374

device-2015-07-26-154212

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
@@ -20,7 +20,6 @@ import android.accounts.Account;
 import android.accounts.AccountManager;
 import android.app.Dialog;
 import android.content.Context;
-import android.content.DialogInterface;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.location.LocationManager;
@@ -32,14 +31,17 @@ import android.os.Build;
 import android.os.Build.VERSION;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
+import android.support.design.widget.TextInputLayout;
 import android.support.v4.app.DialogFragment;
 import android.support.v7.app.AlertDialog;
+import android.text.TextUtils;
 import android.util.DisplayMetrics;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
 import android.widget.AutoCompleteTextView;
+import android.widget.Button;
 import android.widget.EditText;
 
 import org.gdg.frisbee.android.BuildConfig;
@@ -71,8 +73,17 @@ public class FeedbackFragment extends DialogFragment {
     private static final String PROPERTY_APP_VERSION_CODE = "App Version Code";
 //    private static final String POWERED_BY_DOORBELL_TEXT = "Powered by <a href=\"https://doorbell.io\">Doorbell.io</a>";
 
-    @Bind(R.id.feedback_message_text) EditText mMessageField;
-    @Bind(R.id.feedback_email_text) AutoCompleteTextView mEmailField;
+    @Bind(R.id.feedback_message_text)
+    EditText mMessageField;
+
+    @Bind(R.id.feedback_email_text)
+    AutoCompleteTextView mEmailField;
+
+    @Bind(R.id.feedback_message_text_layout)
+    TextInputLayout mLayoutMessage;
+
+    @Bind(R.id.feedback_email_text_layout)
+    TextInputLayout mLayoutEmail;
 
     private JSONObject mProperties;
     private DoorbellApi mApi;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
```diff
@@ -71,8 +73,17 @@
...
-    @Bind(R.id.feedback_message_text) EditText mMessageField;
-    @Bind(R.id.feedback_email_text) AutoCompleteTextView mEmailField;
+    @Bind(R.id.feedback_message_text)
+    EditText mMessageField;
+
+    @Bind(R.id.feedback_email_text)
+    AutoCompleteTextView mEmailField;
+
+    @Bind(R.id.feedback_message_text_layout)
+    TextInputLayout mLayoutMessage;
```
friedger: :+1:

friedger: But where is it defined?

PareshMayani: @friedger it's already defined in dialog feedback layout https://github.com/gdg-x/frisbee/blob/develop/app/src/main/res/layout/dialog_feedback.xml

friedger: I see.
Probably we can get rid of the binding of the EditTexts and use `getEditText`.

friedger:
merging now.

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
PareshMayani: @friedger yes we can!

PareshMayani:
@friedger thanks :)

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
```diff
@@ -105,24 +116,66 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
+            isValid = true;
+            mLayoutMessage.setErrorEnabled(false);
+        } else {
+            isValid = false;
+            mLayoutMessage.setErrorEnabled(true);
+            mLayoutMessage.setError(getString(R.string.feedback_message_required));
+        }
+
+        if (!TextUtils.isEmpty(strEmail) && android.util.Patterns.EMAIL_ADDRESS.matcher(strEmail).matches()) {
+            isValid = true;
```
tasomaniac: We have a problem here. If the message is invalid but email is valid it will override and return valid as a result.

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
```diff
@@ -105,24 +116,66 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
+            mLayoutEmail.setErrorEnabled(false);
+        } else {
+            isValid = false;
+            mLayoutMessage.setErrorEnabled(true);
+            mLayoutEmail.setError(getString(R.string.feedback_invalid_email));
```
tasomaniac: Can't these error messages be set in XML?

app/src/main/res/values/strings.xml
```diff
@@ -172,4 +172,6 @@
     <string name="cd_navdrawer_user_picture">User Profile Image</string>
     <string name="title_signing_needed">Sign-in Needed</string>
     <string name="signin">Sign-in</string>
+    <string name="feedback_message_required">Message required</string>
+    <string name="feedback_invalid_email">Input valid email address</string>
```
tasomaniac: "Email is not valid" would be better.

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
PareshMayani: @tasomaniac there is an attribute for enabling error but not for setting an error message!

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
PareshMayani: @tasomaniac ooops my bad! fixing it up and sending PR in a while.

tasomaniac: I think `isValid = isValid && true;` will help.

PareshMayani: Yes it will help and we can also take two different variables if memory is
not the concern but I really wanted to have a single variable to accomplish
this task!

app/src/main/res/values/strings.xml
Splaktar: :+1:

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
tasomaniac: There is no need for 2 variables.

app/src/main/res/values/strings.xml
tasomaniac: You are a native speaker @Splaktar 
Please help us with the sentences in the app.

PareshMayani: @tasomaniac as far as I know, "Input valid email address" is an instruction to the user in a polite way, where as "Email is not valid" is like a slapping user and saying that hey input valid one :) Rest is up to any of the native speaker :)

Splaktar: Well, `Email is not valid` let's the user know that what they entered is invalid. Where 'Input valid email address` is asking them to do something that they think is already done (and something that seems to be asked already by the input and floating label).

PareshMayani: @Splaktar thanks, will make this change right way.

PareshMayani: BTW, what should we do here? It's just a single string change (for english) but we have localizations implemented so would require string values for other languages.

tasomaniac: It is a tough problem. It already happened for the Lead messages wording changes. 
The translators on crowdin sees those changed strings untranslated with a quick suggestion with the old translation. 

What we can do is that we can wait for the new translations. Or just use the old ones.

PR #420, opened 2015-07-27
Games submenu

tasomaniac:
Here is the PR that does the latest screenshot from me in the comment in #410 

It basically adds a new submenu. 

And I've made a quick change to `navigateTo()` function. The bundle it receives now prepopulated with the id of a selected navigation item. It will check it on the newly created Activity.


Splaktar:
:+1:

PR #422, opened 2015-07-27
Revert google api

tasomaniac:
I tried to get rid of Google API client before. It was always disturbing me because it is a big dependency and we were initializing an OKHttp Client every time we use it. 

I removed it and used `GoogleApiClient` from GMS. But it required for user to be logged in. 

At the end, I reverted it to old system and improved it a little bit. 
`Plus Client, GapiOKTransport and GsonFactory` is stored in App class.
Please tell me if I am wrong. `GoogleApiClient` may be lightweight but `OKHttp` is not. It is a good practice to use only one instanse of OKHttp if it is possible. I even would like to extract an OKHttp instance and use it in Plus Client, Picasso and Retrofit. 

And there was a problem I caused in `getPerson` function. It was not getting the cached object. It is fixed in the last commit.


tasomaniac:
I hope this will fix the performance problems and slowness of the latest builds. Please test it and confirm. It looks good to me now.

fnk0:
It looks good to me! It also goes along my #432 on the OkHtttpClient / retrofit refactors that I did! :)

tasomaniac:
I will go ahead and merge this. Sorry guys. This should be merged to ease the development process. I will not be this free in the following weeks.

PR #432, opened 2015-07-28
* Retrofit refactor

fnk0:
* Moved some code from the activities / app that instantiated adapters to their own files so it's easy to find.
* Made a singe instance of a OkClient linked to a single OkHttpClient with a Cache.
* Made the GDG client to be a singleton

Note: I would love some suggestions on this... Let me know what you guys think.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
@@ -25,9 +25,6 @@ public class GdeDirectoryClient {
     private GdeDirectoryClient() {
     }
 
-    // We can make this a singleton to avoid instantiating this twice
-    private static GdeDirectory gdeDirectory;
-
     private static RestAdapter getRestAdapter() {
         return new RestAdapter.Builder()
                 .setEndpoint(API_URL)
@@ -38,9 +35,6 @@ public class GdeDirectoryClient {
     }
 
     public static GdeDirectory getGdeApi() {
-        if (gdeDirectory == null) {
-            gdeDirectory = getRestAdapter().create(GdeDirectory.class);
-        }
-        return gdeDirectory;
+        return getRestAdapter().create(GdeDirectory.class);
     }
 }
--- a/app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
@@ -1,0 +1,46 @@
+package org.gdg.frisbee.android.api;
+
+import com.google.gson.FieldNamingPolicy;
+
+import org.gdg.frisbee.android.BuildConfig;
+import org.gdg.frisbee.android.app.App;
+import org.gdg.frisbee.android.utils.Utils;
+
+import retrofit.RestAdapter;
+import retrofit.converter.GsonConverter;
+
+/**
+ * Created by <a href="mailto:marcusandreog@gmail.com">Marcus Gabilheri</a>
+ *
+ * @author Marcus Gabilheri
+ * @version 1.0
+ * @since 7/28/15.
+ */
+public class GdeDirectoryClient {
+
+    // This variable is meant to be changed as pleased while debugging
+    private static final RestAdapter.LogLevel LOG_LEVEL = RestAdapter.LogLevel.NONE;
+    public static final String API_URL = "https://gde-map.appspot.com";
+
+    private GdeDirectoryClient() {
+    }
+
+    // We can make this a singleton to avoid instantiating this twice
+    private static GdeDirectory gdeDirectory;
+
+    private static RestAdapter getRestAdapter() {
+        return new RestAdapter.Builder()
+                .setEndpoint(API_URL)
+                .setClient(App.getInstance().getRetrofitClient())
+                .setLogLevel(BuildConfig.DEBUG ? LOG_LEVEL : RestAdapter.LogLevel.NONE)
+                .setConverter(new GsonConverter(Utils.getGson(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)))
+                .build();
+    }
+
+    public static GdeDirectory getGdeApi() {
+        if (gdeDirectory == null) {
+            gdeDirectory = getRestAdapter().create(GdeDirectory.class);
+        }
+        return gdeDirectory;
+    }
+}
--- a/app/src/main/java/org/gdg/frisbee/android/api/GdgXHubClient.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/GdgXHubClient.java
@@ -1,0 +1,40 @@
+package org.gdg.frisbee.android.api;
+
+import com.google.gson.FieldNamingPolicy;
+
+import org.gdg.frisbee.android.BuildConfig;
+import org.gdg.frisbee.android.api.deserializer.ZuluDateTimeDeserializer;
+import org.gdg.frisbee.android.app.App;
+import org.gdg.frisbee.android.utils.Utils;
+
+import retrofit.RestAdapter;
+import retrofit.converter.GsonConverter;
+
+/**
+ * Created by <a href="mailto:marcusandreog@gmail.com">Marcus Gabilheri</a>
+ *
+ * @author Marcus Gabilheri
+ * @version 1.0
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
```diff
@@ -0,0 +1,46 @@
...
+
+    // This variable is meant to be changed as pleased while debugging
+    private static final RestAdapter.LogLevel LOG_LEVEL = RestAdapter.LogLevel.NONE;
+    public static final String API_URL = "https://gde-map.appspot.com";
+
+    private GdeDirectoryClient() {
+    }
+
+    // We can make this a singleton to avoid instantiating this twice
+    private static GdeDirectory gdeDirectory;
```
tasomaniac: I don't think we need static instances here.

app/src/main/java/org/gdg/frisbee/android/api/OkClientProvider.java
```diff
@@ -0,0 +1,45 @@
...
+    private static OkClient okClient;
+
+    private OkClientProvider() {
+    }
+
+    public static OkHttpClient getOkHttpClient(Context context) {
+        OkHttpClient okHttpClient = new OkHttpClient();
+        // Install an HTTP cache in the application cache directory.
+        File cacheDir = new File(context.getApplicationContext().getCacheDir(), "http");
+        Cache cache = new Cache(cacheDir, DISK_CACHE_SIZE);
```
tasomaniac: Why the default cache implementation is not necessary? 

I think just like we store `OkClient` here, we should have only 1 `OkHttpClient` instance as well. 
In one of my current PRs, I created OkHttpClient and used same client within Picasso and Plus Client's HttpTransporter.

app/src/main/java/org/gdg/frisbee/android/api/GdgXHubClient.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private GdgXHubClient() {
+    }
+
+    private static RestAdapter getRestAdapter() {
+        return new RestAdapter.Builder()
+                .setEndpoint(GdgXHub.BASE_URL)
+                .setLogLevel(BuildConfig.DEBUG ? LOG_LEVEL : RestAdapter.LogLevel.NONE)
+                .setConverter(new GsonConverter(Utils.getGson(FieldNamingPolicy.IDENTITY, new ZuluDateTimeDeserializer())))
+                .setClient(App.getInstance().getRetrofitClient())
```
tasomaniac: What if we only have 1 instance of `OkHttpClient` which is the bigger object I suppose. 
Then we can create `new OkClient(App.getInstance().getOkHttpClient())` every time we use it. As far as I know `OkClient` is a wrapper that has static functions to do the request with the given `OkHttpClient`

What do you think?

app/src/main/java/org/gdg/frisbee/android/api/OkClientProvider.java
fnk0: The cache I follow [this recipe from square](https://github.com/square/okhttp/wiki/Recipes#response-caching)
I had it that way previously.We can do that as well. My taught in not making that static is that we are just requesting the OkHttpClient through the OkClient and that one is a singleton making the call to that getOkHttpClient() just once. But we can potentially have both of them be singletons. I'll make the change.

tasomaniac: I am thinking to use OkHttpClient in other places.

fnk0: Ok! Sounds good. So let's see if I understood it. Since the OkClient() is just a wrapper for the OkHttpClient. Should we have just the OkHttpClient be a wrapper and them when we set the client for a RestAdapter we do something like this:
```java
restAdapter.setClient(new OkClient(App.getInstance().getOkHttpClient());
```

Does that sounds good? Or should we have a separate singleton to hold a single OkClient?

app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
fnk0: Any particular reason why we have the other 2 clients as singletons and not have this one?

tasomaniac: The comment is actually for all of them. Since the App object is already a singleton. And we are accessing these using that. I think these would be better if they are not singleton.

app/src/main/java/org/gdg/frisbee/android/api/OkClientProvider.java
tasomaniac: I was exactly thinking about the same. We definitely need an only one instance of OkHttpClient. 
We may clone it though. Cloning OkHttpClient will create a new instance but it will use the same cache object.

app/src/main/java/org/gdg/frisbee/android/api/GdeDirectoryClient.java
fnk0: Makes sense. So just expose those methods and let the App object handle the management of the instances?

tasomaniac: Yes totally. It would be really cool if we could use Dagger to inject
these. Maybe we can in the future.

tasomaniac: If you can do it. I will merge this and rebase my branch on top of this. It
will be a really good improvement.

app/src/main/java/org/gdg/frisbee/android/api/OkClientProvider.java
fnk0: Why not just have one instance as a singleton? Jake Wharton always says on his talks that OkHttpClient is design to be used as a singleton. 
Here's a reference where he mentions it (https://plus.google.com/118239425803358296962/posts/5nzAvPaitHu).

He also always uses OkHttpClient as singleton on the apps he open sources such as u2020.

tasomaniac: I think it should be a singleton. But that doesn't mean that it should be
static. There should one be one instance. And that would be an instance in
App class. Or we can remove it from the app class and have a static
singleton instance. But accessing it through App instance and making it a
static seems too much to me.

fnk0: Ok! Understand now. I will make it not static and have the App class hold a reference to a single instance!

tasomaniac: That was just a suggestion.  What do you think? 

When I look how Dagger is doing it. It has a (not static) Component instance inside App object. And all of the objects provided are not static inside that component whether if it is singleton or not.

fnk0: Sounds like a great idea! I will make those chances when I get home from work and commit to this PR :octopus:

fnk0: I think now it looks the way you mentioned. The references for that are in the Application object and in the ClientProviders there's no static reference to none of them.

app/src/main/java/org/gdg/frisbee/android/api/OkClientProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public final class OkClientProvider {
+
+    // Cache size for the OkHttpClient
+    private static final int DISK_CACHE_SIZE = 50 * 1024 * 1024; // 50MB
+
+    private OkClientProvider() {
+    }
+
+    public static OkHttpClient getOkHttpClient(Context context) {
```
tasomaniac: But these are still static. Why are they still static?

tasomaniac: Sorry these are not functions. Never mind.

tasomaniac:
Some function names are changed and manually merged.

fnk0:
@tasomaniac I like the factory name better also! Thanks for changing it!

tasomaniac:
Thank you for the contribution. I kept your commits when I was merging. Keep them coming.

PR #433, opened 2015-07-28
Move LeadMessage class into the model package

PareshMayani:
LeadMessage class was in Lead package but it should be in model package. Moved it there.


tasomaniac:
Henrique did it like that. Not only that but also other classes like that are in their own packages.

PareshMayani:
I just found it and felt that it should be in model package so moved it. Hope I haven't made a mistake!

tasomaniac:
I think it should be in the model package as well. But the thing is lots of
things are moved in Henrique s pr. And moving them again would not make
sense.

PareshMayani:
@tasomaniac surprisingly it's not created/committed by Henrique but by Friedger :) 
lead folder history

tasomaniac:
Oh ok. :smile:

PR #436, opened 2015-07-29
Chapter swap fix

### LINKED ISSUE #423: Chapter swap in main page not working.
It simply does not reload the pager fragment s

tasomaniac:
Some refactoring in the ChapterAdapter and how selected chapter is handled in MainActivity. 

Bug fix related to #423 ViewPager was not updating itself, because of the getItemPosition() not overriden. That's my fault. 
Wrote tests for chapter swap and orientation change. 

No UI changes. 

Fixes #423

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
@@ -273,7 +273,7 @@ public class MainActivity extends GdgNavDrawerActivity {
         mViewPagerAdapter = new ChapterFragmentPagerAdapter(this,
                 getSupportFragmentManager(), selectedChapterId);
         mViewPager.setAdapter(mViewPagerAdapter);
-        mViewPager.setOffscreenPageLimit(2);
+        mViewPager.setOffscreenPageLimit(3);
         updateSelectionfor(selectedChapterId);
 
         mTabLayout.setupWithViewPager(mViewPager);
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -387,8 +359,9 @@ private void initSpinner() {
             public void onItemSelected(final AdapterView<?> parent, final View view, final int position, final long id) {
 
                 Chapter selectedChapter = mChapterAdapter.getItem(position);
-                updateSelectionfor(selectedChapter);
+                updateSelectionfor(selectedChapter.getGplusId());
```
fnk0: I wonder if we should keep a reference to the previous selected chapter and in case we the user for some reason reselects the same chapter we avoid doing a page reload again.

fnk0:
It looks good to me. There's just one thing that is sort of bothering me.
The LeadFragment is always the same right? Right now we are only keeping 2 items offscreen and this activity has 4 fragment pages. If you swipe all the way to the end and them back to news and back to LeadFragment it causes NewsFragment to leak. We also have to do all the calls to load the news again. 

And correct me if I'm wrong... that LeadFragment only appears if the person signed in is a organizer right? 

We could potentially keep 3 items offscreen avoiding making a NewInstance() of those 2 fragments every single time.

tasomaniac:
You are definitely right. This is something we forget when we add LeadsFragment. Thank you very much.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
tasomaniac: You mean keeping a reference to the previous chapter's id or the did you mean caching the whole Fragment somehow. 
I don't think reloading is a problem. We should do the caching better though. It is a different problem.

fnk0:
Looks good now!

Issue #269, opened 2015-03-02
Redesign InfoFragment


tasomaniac:
InfoFragment now uses cards in portrait layout. 
It still uses old 9patch cards in landscape. 

Both of them are ugly and should be fixed.

mauimauer:
Also paddings/margins seem a little screwed up.

mauimauer:
![screenshot_2015-03-02-23-03-27](https://cloud.githubusercontent.com/assets/555852/6452893/b0e47ee6-c130-11e4-9ebb-c3686607a710.png)
Nexus 6

tasomaniac:
Yep. That is the problem with all of the new card view usages in the app.

PareshMayani:
@tasomaniac @mauimauer that padding issue has already been resolved in my pull request https://github.com/gdg-x/frisbee/pull/308 :)

tasomaniac:
Landscape should be fixed. :) And keyline rythm should be followed. 

- All the body text should use keyline1 padding. 
- Organizer photos should have keyline1 padding. 
- Organizer names should have keyline2 padding from the left.

tasomaniac:
I will close this issue since it is way much better now. 
Closed via #437

PR #437, opened 2015-07-30
Info Fragment

fnk0:
Fix for #269 

* Separated the cards for the fragment_info into it’s own separate files
* Modified the cards from fragment_info to follow the same design as
the cards in the GDE info fragment.

### Screenshots

portrait

landscape


tasomaniac:
Do you think we should make the profile images circular? 
:)

PareshMayani:
@tasomaniac yes rounded one would look better!

fnk0:
@tasomaniac I'm always in favor of rounded images!! I'll do the change and push a commit in a little bit!

tasomaniac:
Although @friedger s work contains this, I will merge this because it is small and nice improvement. 
Thank you for the contribution. 
@friedger you might have to rebase your branch. There will be conflicts.

PR #441, opened 2015-08-01
Fix small drawer bug in ArrowTaggedActivity

fnk0:
* Fixed small bug on the nav drawer selected item related that if a user is in the ArrowTaggedActivity the selected item in the drawer would be Home rather than Arrow


tasomaniac:
Instead can you make the ArrowTaggedActivity.java to extends normal Activity, not the NavDrawer one. And have back button on the top left. What do you think?

fnk0:
@tasomaniac  I like that idea better also. If something doesn't have their place in the drawer it should not have the hamburger icon! I will change that and commit to this PR

PR #443, opened 2015-08-01
refactor: app widget codes.

tasomaniac:
Removed unnecessary things.


tasomaniac:
Although the changes look like big, this PR just removes unused `findChapter`. The code blocks are exactly the same. Only indentation changes. I think we can merge this.

PR #445, opened 2015-08-02
"Infer Nullity" in Android Studio is run.

tasomaniac:
This does not change any behavior. It just adds annotations.

I know this is a big PR. Sorry about that. I've just run Infer Nullity and checked all the changes one by one. While checking fixed couple of warning about nullity. That's all. 

If someone wants to review this, it can wait until we released 2.2

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
@@ -17,6 +17,7 @@
 package org.gdg.frisbee.android.about;
 
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.design.widget.TabLayout;
 import android.support.v4.view.ViewPager;
 import android.view.MenuItem;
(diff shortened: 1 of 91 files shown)


stefanhoth:
At work we usually go with `@NonNull` as the default expectation. This way we don't need to plaster the whole codebase with this annotation and only mark the rare cases of `@Nullable` which are much more important to handle.

I wanted to raise this as a potential project agreement to minimize the "noise" in the codebase.

tasomaniac:
How do you do that? It actually makes sense. 
When we have `@NonNull` Android Studio warns for potential NullPointerExceptions. Is there an option to say that act like if there is `@NonNull` if it is not annotated? If there is one, would it be possible to integrate it to every contributor of the project.

stefanhoth:
@tasomaniac No, we decided that we only let AS warn us about potential NPEs with `@Nullable` returns or input values.

If a developer sends a `null` value to a field that is not annotated the app is supposed to crash quickly, showing the developer what he/she did wrong. :wink:

tasomaniac:
What should we do about this PR? Should I close it, should I change to have only @Nullable annotations. 

@stefanhoth What you are suggesting really makes sense. But it makes sense more within a company where everyone knows the rule. If we have @NonNull annotations, it will be safer for new contributors.

stefanhoth:
@tasomaniac True but could also just introduce checkstyle/pmd/findbugs to check for those issues. Not sure if that would be too strict but would help with a consistent quality level of contributions.

tasomaniac:
I think that would be really nice. I hesitate to integrate pmd/findbugs because if I did, I may have to fix lots of bugs that it reports just to build the project. Initial integration may be really hard.

friedger:
@tasomaniac I agree with @stefanhoth . I close the PR now.

PR #446, opened 2015-08-02
Fix: Initial Load Problem and GDE List Problems

### LINKED ISSUE #439: GDE list is not showing on release build
Gde page is ok on debug builds but it is empty on release builds. It maybe because of a proguard rule problem. We should look into it before full release. 

I won't be able to look at it. 
If someone can change the release keystore to debug temporarily and test it. It would be great.

### LINKED ISSUE #440: Chapters load problem
On the first open it doesn't load the chapters somehow. It downloads them and put them into cache but never displays them. 
When you open the app one more time it displays chapter spinner through cache.

tasomaniac:
Fixes #439 and fixes #440

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/app/App.java
+++ b/app/src/main/java/org/gdg/frisbee/android/app/App.java
@@ -284,6 +284,7 @@ public class App extends Application implements LocationListener {
         return mTracker;
     }
 
+    @NonNull
     public ModelCache getModelCache() {
         if (mModelCache == null) {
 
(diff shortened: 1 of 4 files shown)


Splaktar:
I think that I follow most of this and it looks good. But can you explain a little more of what you are doing with the removal of the WeakReferences and why they don't work for this?

tasomaniac:
This `ModelCache` `AsyncTaks`s were leaking `Activity` `Context` a lot. When you scroll through GDE tabs, it was really easy to get Leak Canary leak. 

I made the `AsyncTak`s classes static before. While I was doing that I took extra precaution not to leak any `Context` and made the listener object `WeakReference`. I don't know why but although it works on Debug builds fine, I was getting null when the code tries to reach to the listener objects. 

I removed WeakReference and GDE page doesn't leak anything. I think the initial `WeakReference` usage was too much.

Splaktar:
OK LGTM

PR #431, opened 2015-07-28
* ArrowTaggedActivity Improvements / fixes

fnk0:
* Changed the ArrowTaggedActivity from ListView to RecyclerView
* Made the list items follow more of the design guidelines and changed cards to simple list items
* Fixed a bug that would keep making the list get populated if user hit home button and them came back to the app
* Reduced 1 layer of Overdraw by moving from CardView to a normal list
* Improved overall look of the list

#### Before (Cards with Overdraw)
before

#### After (List items without overdraw)
after

after_nogpu

PR #455, opened 2015-08-03
update(arrow): Clean up Arrow code

Splaktar:
- Extract classes and methods out of ArrowTaggedActivity to make the code more manageable.
- This included renaming a number of variables to make them more obvious.
- Fix Accessibility issues with images on ArrowActivity.
- Fix Exception related to GoogleApiClient not being connected when calling getEncryptedMessage() in ArrowActivity.
- Fix warning about possible NPE in createNdefMessage().
- Log exceptions with Timber in ArrowActivity.
- Some of these changes are based on feedback in #431 

There is still a resource leak happening here that I haven't had time to track down, but I don't believe that it is something new introduced by these changes. Please let me know if you can see it.

    org.gdg.frisbee.android.debug E/DataBuffer﹕ Internal data leak within a DataBuffer object detected!  Be sure to explicitly call release() on all DataBuffer extending objects when you are done with them. (com.google.android.gms.appstate.AppStateBuffer@3e988)
    org.gdg.frisbee.android.debug E/DataBuffer﹕ Internal data leak within a DataBuffer object detected!  Be sure to explicitly call release() on all DataBuffer extending objects when you are done with them. (com.google.android.gms.common.data.zzd@e15702f)

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/arrow/ArrowActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/arrow/ArrowActivity.java
@@ -65,6 +65,7 @@ import java.util.HashSet;
 import java.util.Set;
 
 import butterknife.Bind;
+import timber.log.Timber;
 
 public class ArrowActivity extends GdgNavDrawerActivity {
 
@@ -212,7 +213,6 @@ public class ArrowActivity extends GdgNavDrawerActivity {
     }
 
     private void taggedPerson(String msg) {
-
         try {
             String decrypted = CryptoUtils.decrypt(Const.ARROW_K, msg);
 
@@ -235,7 +235,7 @@ public class ArrowActivity extends GdgNavDrawerActivity {
             }
 
         } catch (Exception e) {
-            e.printStackTrace();  //To change body of catch statement use File | Settings | File Templates.
+            Timber.e(e, "Error while trying to tag person");
         }
     }
 
@@ -313,10 +313,14 @@ public class ArrowActivity extends GdgNavDrawerActivity {
 
 
     private String getEncryptedMessage() throws Exception {
-        return CryptoUtils.encrypt(Const.ARROW_K, 
-                Plus.PeopleApi.getCurrentPerson(getGoogleApiClient()).getId()
-                + ID_SPLIT_CHAR
-                + getNow());
+        if (getGoogleApiClient().isConnected()) {
+            return CryptoUtils.encrypt(Const.ARROW_K,
+                    Plus.PeopleApi.getCurrentPerson(getGoogleApiClient()).getId()
+                            + ID_SPLIT_CHAR
+                            + getNow());
+        } else {
+            return null;
+        }
     }
 
     @Override
@@ -388,7 +392,7 @@ public class ArrowActivity extends GdgNavDrawerActivity {
 
                 mHandler.postDelayed(updateQrCode, 60000);
             } catch (Exception e) {
-                e.printStackTrace();
+                Timber.e(e, "Error while trying to update QR code");
             }
         }
     };
@@ -405,7 +409,9 @@ public class ArrowActivity extends GdgNavDrawerActivity {
         }
     }
 
-    private class NfcArrowHandler extends BaseArrowHandler implements NfcAdapter.OnNdefPushCompleteCallback, NfcAdapter.CreateNdefMessageCallback {
+    private class NfcArrowHandler extends BaseArrowHandler
+            implements NfcAdapter.OnNdefPushCompleteCallback, NfcAdapter.CreateNdefMessageCallback {
+
         public void enablePush() {
             mNfcAdapter.setNdefPushMessageCallback(this, ArrowActivity.this);
             mNfcAdapter.setOnNdefPushCompleteCallback(this, ArrowActivity.this);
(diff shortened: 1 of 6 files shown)


fnk0:
It looks good to me. Can you do one quick check tough. Go to that screen and them hit home button and come back to the app. And see if the list gets duplicated or not.

app/src/main/java/org/gdg/frisbee/android/arrow/ArrowTaggedActivity.java
```diff
@@ -87,78 +76,80 @@ public void onCreate(Bundle savedInstanceState) {
...
+                            public void failure(final RetrofitError error) {
+                                Timber.e(error, "Error");
+                            }
+                        });
                     }
                 });
+    }
+
+    private void loadChapterOrganizers(Directory directory) {
+        if (Strings.isNullOrEmpty(serializedOrganizers)) {
```
tasomaniac: Can you use TextUtils.isEmpty() for this. Thank you :small_blue_diamond:

app/src/main/java/org/gdg/frisbee/android/arrow/Organizer.java
```diff
@@ -0,0 +1,42 @@
+package org.gdg.frisbee.android.arrow;
+
+import com.google.android.gms.plus.model.people.Person;
+
+public class Organizer {
```
tasomaniac: We have lots of POJOs like this. It can be easily confused with others I think. Even if it is in the arrow package. What do you think?

app/src/main/res/values/strings.xml
```diff
@@ -173,4 +173,6 @@
     <string name="feedback_invalid_email">Input valid email address</string>
     <string name="drawer_subheader_games">Games</string>
     <string name="fork_me">Fork me</string>
+    <string name="arrowTagImage">Arrow Tagging Game</string>
```
tasomaniac: What do you think about putting cd_ prefix to content description strings.

app/src/main/java/org/gdg/frisbee/android/arrow/OrganizerAdapter.java
```diff
@@ -0,0 +1,86 @@
...
+import android.widget.TextView;
+
+import org.gdg.frisbee.android.R;
+import org.gdg.frisbee.android.app.App;
+import org.gdg.frisbee.android.view.BitmapBorderTransformation;
+
+import butterknife.Bind;
+import butterknife.ButterKnife;
+
+public class OrganizerAdapter extends RecyclerView.Adapter<OrganizerAdapter.ViewHolder> {
```
tasomaniac: Thank you for doing this :smile:

app/src/main/java/org/gdg/frisbee/android/arrow/ArrowActivity.java
```diff
@@ -313,10 +313,14 @@ public void showNoNfc() {
 
 
     private String getEncryptedMessage() throws Exception {
-        return CryptoUtils.encrypt(Const.ARROW_K,
```
tasomaniac: Since this returns null now, it should be null checked. I believe you did it in only 1 place. It is also used in QR generation. 
It would help if you use @Nullable annotation on return statement.

tasomaniac:
As far as I see, @fnk0 's concern is valid. The old bug where it tries to fetch it on every app being foreground is present right now. 
An easy way to fix that is to move the networking code from onStart to onCreate.

app/src/main/java/org/gdg/frisbee/android/arrow/ArrowTaggedActivity.java
Splaktar: Done.

app/src/main/java/org/gdg/frisbee/android/arrow/Organizer.java
Splaktar: I don't see another `Organizer` class in the project. Maybe I don't understand your point?

app/src/main/res/values/strings.xml
Splaktar: Never heard of something like that. I can add it though. I don't see any other strings with the `cd_` prefix.

app/src/main/java/org/gdg/frisbee/android/arrow/ArrowActivity.java
Splaktar: Done.

Splaktar:
Fixed old issue of duplicating the list of tagged organizers when returning to the activity via `onStart`.

app/src/main/res/values/strings.xml
tasomaniac: No we use it in couple of places. It helps the translator.

app/src/main/java/org/gdg/frisbee/android/arrow/Organizer.java
tasomaniac: My bad, I thought we have. Nevermind

app/src/main/java/org/gdg/frisbee/android/arrow/ArrowActivity.java
```diff
@@ -311,12 +312,16 @@ public void showNoNfc() {
         Toast.makeText(this, R.string.no_nfc_use_qr_scanner, Toast.LENGTH_LONG).show();
     }
 
-
+    @Nullable
     private String getEncryptedMessage() throws Exception {
```
tasomaniac: I think there should be another null check in `updateQrCode` runnable.

Splaktar: This code below isn't enough (line 376)?

                if (TextUtils.isEmpty(encryptedMessage)) {
                    return;
                }

PR #456, opened 2015-08-03
fix(events): Improve sorting of chapter events

### LINKED ISSUE #311: Event listing with upcoming and past separation
Currently events are getting displayed based on the feed we are getting, but there should be 2 sections:
1) Upcoming
2) Past

fnk0:
* Fixes the sorting of the events in EventListFragment. 

Sorry about the revert and recommit, etc.. the other commit was getting in the way! 
Having some branch problems, srry! 

Fixes #311

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
@@ -20,6 +20,7 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.design.widget.Snackbar;
+import android.util.Pair;
 import android.view.ContextMenu;
 import android.view.LayoutInflater;
 import android.view.MenuItem;
@@ -34,8 +35,12 @@ import org.gdg.frisbee.android.api.model.SimpleEvent;
 import org.gdg.frisbee.android.common.GdgListFragment;
 import org.gdg.frisbee.android.event.EventActivity;
 import org.gdg.frisbee.android.view.ColoredSnackBar;
+import org.joda.time.DateTime;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 import butterknife.ButterKnife;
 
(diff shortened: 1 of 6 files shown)


tasomaniac:
Instead of only 1 commit, we will merge 5 commits. It will look ugly in the history. 
I can manually fix it and merge only one of them or can you fix it? :smile: 

You can create a new branch and cherry pick the right commit or you can rebase interactively and squash these into 1 commit.

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
```diff
@@ -150,6 +155,44 @@ public void addEventToCalendar(SimpleEvent event) {
...
+        DateTime now = DateTime.now();
+
+        for (SimpleEvent event : events) {
+            if (event.getStart().isBefore(now)) {
+                past.add(event);
+            } else {
+                upcoming.add(event);
+            }
+        }
+        Collections.reverse(past);
```
tasomaniac: I think we should reverse the upcoming list. Am I wrong? I am not sure.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeriesFragment.java
```diff
@@ -108,6 +108,7 @@ EventAdapter createEventAdapter() {
         return new EventAdapter(getActivity(), mTaggedEventSeries.getDefaultIconResId());
     }
 
+    @SuppressWarnings("unchecked")
```
tasomaniac: These warnings are dangerous actually and we have to have something to fix that. These caused crashes in the past. 
Can you remove them? :smile:

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
```diff
@@ -150,6 +155,44 @@ public void addEventToCalendar(SimpleEvent event) {
...
+
+    /**
+     * Split a events list into Upcoming / Past events
+     *
+     * @param events
+     *      The events list that needs to be sorted
+     * @return
+     *      A pair containting the split list into Upcoming / Past events
+     */
+    protected Pair<List<SimpleEvent>, List<SimpleEvent>> splitEventsList(Collection<? extends SimpleEvent> events) {
```
tasomaniac: I think we can make this private.

tasomaniac:
Thank you for the work. It looks great. 
I can also be done with a custom comparator where you compare things and decide which one to go first or not. 
You can look at `ChapterComparator` of ours.

fnk0:
@tasomaniac I did a custom comparator at first but the problem is that there's a different behavior for each of them. That's why we reverse only the Past events list. If I used a custom comparator by date them we would have the same behavior even if the date was on the past. I might try later on making it all in a single Comparator and see if works. I only tried the comparator thing for about 10 minutes.

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
fnk0: That is in the abstract fragment and I call it from 2 of the fragments that extends EventListFragment. That's why I made it protected.

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
fnk0: No. This actually sorts like this. The Upcoming event list is sorted in ASC order as we want the ones that are the closest to the date we are first and the Past events order is sorted in DESC order as we want the most recents one on top.

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
tasomaniac: No I think you are calling the above in the other `splitEventsAndAddToAdapter ` Fragments. This one is not called.

fnk0: Got it! I can make it private them. And if in the future we just need to split the events without adding to the adapter we can change it to protected.

tasomaniac:
1 comparator was just an idea. I think it can be done. You can control the behavior based on if the event is past or not. 
But I now think that it is not necessary because we have to split the Array into 2 anyways to display in 2 sections in the future.

app/src/main/java/org/gdg/frisbee/android/eventseries/EventListFragment.java
tasomaniac: You are absolutely right. I was confused. Thank you for the explanation.

tasomaniac:
Rebased, deleted wrong commits, changed code review changes and manually merged.

Issue #457, opened 2015-08-03
AchievementHandler should be redesigned. 


tasomaniac:
Here is a crash occured in AchievementHandler. It is not easy to fix. It should be redesigned. 

Here is the public link to crash. 
http://crashes.to/s/06d4aca4397

Splaktar:
java.lang.IllegalStateException: GoogleApiClient is not connected yet.
       at com.google.android.gms.common.api.zzf.zzb()
       at com.google.android.gms.common.api.zzg.zzb()
       at com.google.android.gms.games.internal.api.AchievementsImpl.getAchievementsIntent()
       at org.gdg.frisbee.android.achievements.AchievementActionHandler$1.run(AchievementActionHandler.java:80)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5254)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)

tasomaniac:
This should be fixed in Release 2.3

PR #458, opened 2015-08-03
Achievement Unlocks are now stored in SharedPreferences.

tasomaniac:
This way we stopped calling Achievement unlocking in every Activity start.
This also stops leaking Activities I hope.
I will also help #457 but will not fix it completely.


Splaktar:
LGTM

PR #459, opened 2015-08-04
id and url fields are added to the Plus Api request.

### LINKED ISSUE #452: InfoFragment organizer profile click open wrong link.
It opens http://plus.google.com/null/posts

tasomaniac:
getUrl() used instead of getId() and manually creating Url.
If getUrl() is empty, nothing is done.

Fixes #452


stefanhoth:
This seems to be a good fix but it would have been nice if the description would have been a bit more descriptive. :wink:

stefanhoth:
Ok, let's :ship: it!

PR #461, opened 2015-08-10
refactor: Plus Api getPerson duplicate code is removed.

tasomaniac:
Just a duplication removal. No functionality change, no UI change.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/api/PlusPersonDownloader.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/PlusPersonDownloader.java
@@ -19,7 +19,7 @@ import java.util.regex.Pattern;
 import timber.log.Timber;
 
 /**
- * Created by tasomaniac on 27/7/15.
+ * Created by Said Tahsin Dane on 27/7/15.
  */
 public class PlusPersonDownloader implements Interceptor {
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/gdg/frisbee/android/api/PlusPersonDownloader.java
```diff
@@ -51,7 +51,12 @@ public Response intercept(Chain chain) throws IOException {
     }
 
     @Nullable
-    public Person getPersonSync(final String gplusId) {
+    public static Person getPersonSync(final String gplusId) {
```
friedger: Do we use more than this instance of the gplus client? If yes, probably it would be good to add this to the method name.

tasomaniac: I didn't understand the question. :( Sorry.

friedger: It could be names `getPersonSyncWithAppGplusClientInstance`. This makes sense if there are more instances.

tasomaniac: Plus Client is not a lightweight object as far as I understand. It requires http transport which requires okhttp and gson. 
I moved all PlusClient instances to one instance in App class.

tasomaniac: Eventually I actually want to remove static and make this private function of this class. It will only be accessed here. It requires some work.

PR #462, opened 2015-08-11
refactor: Removed unused dependency Otto

tasomaniac:
OMG! We were not using Otto at all. Dependency removed.


tasomaniac:
Thank you @friedger

PR #463, opened 2015-08-11
Remove some unused code

tasomaniac:
No UI change, no implementation change.


tasomaniac:
Wow. 21 files are removed in total.

app/src/main/java/org/gdg/frisbee/android/fragment/SwipeRefreshRecyclerViewFragment.java
```diff
@@ -165,8 +165,8 @@ private boolean canListViewScrollUp(RecyclerView recyclerView) {
                     return position != 0;
                 } else if (layoutManager instanceof StaggeredGridLayoutManager) {
                     int[] positions = ((StaggeredGridLayoutManager) layoutManager).findFirstCompletelyVisibleItemPositions(null);
-                    for (int i = 0; i < positions.length; i++) {
-                        if (positions[i] == 0) {
+                    for (int position : positions) {
```
friedger: Does it reduce performance?

friedger:
There is too much in the PR. 

I wouldn't merge it because of the changes to the translations. This will be reverted at the next import?

app/src/main/java/org/gdg/frisbee/android/fragment/SwipeRefreshRecyclerViewFragment.java
tasomaniac: Android Studio suggested it. I think in fact it improves it a little bit because in the previos one `position.length` is called every time.

tasomaniac:
@friedger your concern about the translation is totally valid. I forgot to mention it. I refactor them and upload them to Crowdin. They are now all like Android Studio likes. When someone do a translation, they may for example change *…* into *...*. When this happends Crowdin actually warn the translator that there is character in the original text that it is missing in your translation etc. 

And for the other changes. I just run lint's unused resources check and deleted them. They are really really unused.

PR #465, opened 2015-08-11
Made lots of declaration access weaker.

tasomaniac:
Thanks to lint.
Again, there should be no functional change.


friedger:
Does it reduce testability?

tasomaniac:
It may reduce testability but I don't think it is a problem. If we write additional test, and toss a private/protected class/function, we can change its visibility and thats all. Thank you for the merge.

PR #468, opened 2015-08-14
GoogleApiClient hopeful memory leak fix. 

tasomaniac:
Listeners are registered in onStart and unregistered in onStop


stefanhoth:
I love the title.

tasomaniac:
I will be honest. :) It actually doesn't fix. But it is an improvement.

tasomaniac:
And I like the gif :)

friedger:
I love the change, but don't have a gif ;-)

PR #475, opened 2015-08-20
dependencies and build tools are updated.

tasomaniac:
targetSdkVersion is still 22 because we need to test Runtime Permission.


tasomaniac:
Once again we hit 65K dex limit. This time it is because of support library v23.

PR #478, opened 2015-09-01
Fix external Intent problem. 

### LINKED ISSUE #477: External Intent problem.
When Google Plus link is opened in the app, it look like this. It is a famous Android problem. In order to solve it we need to add couple of flags into the Intent. 

In the below screenshot, you can see that Google+ app is open with red primaryColor but with our GDG app logo. 

screenshot_2015-08-30-23-44-32

tasomaniac:
Introduced new utility function `createExternalIntent`
It creates and Intent with ACTION_VIEW with appropriate flags and extras.

Fixes #477 

Old screenshot can be found in #477 
It now looks like this. GDG app and Google+ are separate entities now. 
image

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
+++ b/app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
@@ -16,14 +16,17 @@
 
 package org.gdg.frisbee.android.utils;
 
+import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
 import android.net.ConnectivityManager;
+import android.net.Uri;
 import android.os.Build;
 import android.support.annotation.NonNull;
+import android.support.v4.app.ShareCompat;
 import android.util.DisplayMetrics;
 import android.util.Patterns;
 import android.util.TypedValue;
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
```diff
@@ -209,8 +212,18 @@ public static boolean isEmailAddress(String possibleEmail) {
...
 
-
     public static boolean canLaunch(Context context, final Intent viewUrlIntent) {
         return context.getPackageManager().resolveActivity(viewUrlIntent, PackageManager.MATCH_DEFAULT_ONLY) != null;
     }
+
+    @SuppressWarnings("deprecation")
+    public static Intent createExternalIntent(Context context, Uri uri) {
+        Intent intent = new Intent(Intent.ACTION_VIEW, uri);
+        intent.putExtra(ShareCompat.EXTRA_CALLING_PACKAGE, context.getPackageName());
```
tasomaniac: This Extra's are taken from ShareCompat's implementation. The external Activity we open might use this information if they want to using ShareCompat.IntentReader.

app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
```diff
@@ -209,8 +212,18 @@ public static boolean isEmailAddress(String possibleEmail) {
...
+        if (context instanceof Activity) {
+            intent.putExtra(ShareCompat.EXTRA_CALLING_ACTIVITY, ((Activity) context).getComponentName());
+        }
+        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET);
```
tasomaniac: The most important thing is this flag.

friedger:
Would be nice to add a custom chrome cast tab. Maybe out of scope?

tasomaniac:
Definitely. But that should be another issue. 
When opening a Google Plus link I don't think Custom tabs are necessary but for the other things like Help, Wisdom Book, custom tabs would be awesome.

PR #479, opened 2015-09-03
PulseActivity now goes back to Global on back button press.

### LINKED ISSUE #450: Back press on pulse screen
When a country is selected, my first action to go back to global list is to press back button. It closes the activity. We can override back press and make it go back to global list.

tasomaniac:
Fixes #450


app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
```diff
@@ -174,7 +183,7 @@ public void onNothingSelected(final AdapterView<?> parent) {
         });
 
         Collections.sort(mPulseTargets);
-        mPulseTargets.add(0, "Global");
+        mPulseTargets.add(0, PulseFragment.GLOBAL);
```
friedger: :+1:

PR #481, opened 2015-09-03
Fix pulse numbering bug. 

### LINKED ISSUE #451: Pulse numbering bug.
As it can be seen in the screenshots, we have a problem with the numbering of the items. 

screenshot_20150802-234650
screenshot_20150802-234709

tasomaniac:
The problem is because FragmentStatePagerAdapter was aggressively caching the states of its children Fragments. 
Here is the bug reported and it is still open. https://code.google.com/p/android/issues/detail?id=37990

There are couple of ways to fix it using a really really custom FragmentStatePagerAdapter without subclassing it. I don't like those solutions. It is not optimal but recreating the adapter fixes this problem. 

The problem is actually because of the positions calculation in the code. On orientation change it reset and makes another problem. I saved it on config change and now it works perfectly. 

Fixes #451

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
@@ -239,15 +239,7 @@ public class PulseActivity extends GdgNavDrawerActivity implements PulseFragment
 
         @Override
         public Fragment getItem(int position) {
-            switch (position) {
-                case 0:
-                    return PulseFragment.newInstance(0, mSelectedPulseTarget);
-                case 1:
-                    return PulseFragment.newInstance(1, mSelectedPulseTarget);
-                case 2:
-                    return PulseFragment.newInstance(2, mSelectedPulseTarget);
-            }
-            return null;
+            return PulseFragment.newInstance(position, mSelectedPulseTarget);
         }
 
         @Override
--- a/app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
@@ -76,10 +76,7 @@ public class PulseActivity extends GdgNavDrawerActivity implements PulseFragment
         setContentView(R.layout.activity_pulse);
 
         mPulseTargets = new ArrayList<>();
-
         mViewPagerAdapter = new PulsePagerAdapter(this, getSupportFragmentManager());
-        mSpinnerAdapter = new ArrayAdapter<>(this, R.layout.spinner_item_actionbar);
-        mSpinnerAdapter.setDropDownViewResource(R.layout.support_simple_spinner_dropdown_item);
 
         final String selectedPulse = savedInstanceState != null ? savedInstanceState.getString(INSTANCE_STATE_SELECTED_PULSE) : null;
 
@@ -156,6 +153,9 @@ public class PulseActivity extends GdgNavDrawerActivity implements PulseFragment
     }
 
     private void initSpinner(String selectedPulse) {
+        Collections.sort(mPulseTargets);
+        mPulseTargets.add(0, PulseFragment.GLOBAL);
+
         Toolbar toolbar = getActionBarToolbar();
         View spinnerContainer = LayoutInflater.from(this).inflate(R.layout.actionbar_spinner,
                 toolbar, false);
@@ -164,6 +164,11 @@ public class PulseActivity extends GdgNavDrawerActivity implements PulseFragment
         toolbar.addView(spinnerContainer, lp);
 
         mSpinner = (Spinner) spinnerContainer.findViewById(R.id.actionbar_spinner);
+
+        mSpinnerAdapter = new ArrayAdapter<>(this, R.layout.spinner_item_actionbar, mPulseTargets);
+        mSpinnerAdapter.setDropDownViewResource(R.layout.support_simple_spinner_dropdown_item);
+        mSpinner.setAdapter(mSpinnerAdapter);
+
         mSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
             @Override
             public void onItemSelected(final AdapterView<?> parent, final View view, final int position, final long id) {
@@ -184,21 +189,13 @@ public class PulseActivity extends GdgNavDrawerActivity implements PulseFragment
 
     private void refreshSpinner(String selectedPulse) {
 
-        mSpinner.setAdapter(mSpinnerAdapter);
-
-        Collections.sort(mPulseTargets);
-        mPulseTargets.add(0, PulseFragment.GLOBAL);
         if (selectedPulse == null) {
             selectedPulse = mPulseTargets.get(0);
         }
-
-        mSpinnerAdapter.clear();
-        mSpinnerAdapter.addAll(mPulseTargets);
         mSpinner.setSelection(mPulseTargets.indexOf(selectedPulse));
 
         mViewPagerAdapter = new PulsePagerAdapter(this, getSupportFragmentManager());
         mViewPagerAdapter.setSelectedPulseTarget(selectedPulse);
-
         mViewPager.setAdapter(mViewPagerAdapter);
         mTabLayout.setupWithViewPager(mViewPager);
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
```diff
@@ -117,23 +122,16 @@ public void setPulse(final int mode, Pulse pulse) {
...
-                switch (mode) {
-                    case 0:
-                        return (value.getMeetings() - value2.getMeetings()) * -1;
-                    case 1:
-                        return (value.getAttendees() - value2.getAttendees()) * -1;
-                    case 2:
-                        return (value.getPlusMembers() - value2.getPlusMembers()) * -1;
-                }
-                return 0;
+                return entry.getValue().compareTo(mode, entry2.getValue());
```
tasomaniac: compareTo is a function in `PulseEntry`. And it was exact duplicate of this.

PareshMayani:
:+1: awesome contribution @tasomaniac

tasomaniac:
@PareshMayani you already reviewed this right?
Any one up for merge @friedger @Splaktar ?

app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
```diff
@@ -182,16 +179,25 @@ public void onNothingSelected(final AdapterView<?> parent) {
             }
         });
 
+        refreshSpinner(selectedPulse);
+    }
+
+    private void refreshSpinner(String selectedPulse) {
+
+        mSpinner.setAdapter(mSpinnerAdapter);
```
stefanhoth: This line needed to be reset every time? Looks out of scope for this method.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
```diff
@@ -117,23 +123,16 @@ public void setPulse(final int mode, Pulse pulse) {
         mMode = mode;
         mPulse.clear();
         mPulse.addAll(pulse.entrySet());
-        mPosition = new int[mPulse.size()];
+        //Only initialize if the positions are not provided.
+        //They are provided in constructor and coming from savedInstanceState of the Fragment.
+        if (mPositions == null || mPositions.length != mPulse.size()) {
```
stefanhoth: Should this init be down here? It feels this could be done at the beginning of the method or better yet in constructor.

tasomaniac: We don't know the size before. Network operation is done, Pulse object is filled and then this is called. That's where we know the size.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
tasomaniac: You are right. Fixing now.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
stefanhoth: Could you just make `PulseEntry` implement `Comparable` and avoid this here altogether?

stefanhoth:
I reviewed bits of the code but I don't understand enough to assess the change. If you could explain again what you persist, where and why it would help me understand. It'll be best to comment the lines where something is happening.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseActivity.java
```diff
@@ -164,15 +164,12 @@ private void initSpinner(String selectedPulse) {
...
-        mSpinner.setAdapter(mSpinnerAdapter);
         mSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
             @Override
             public void onItemSelected(final AdapterView<?> parent, final View view, final int position, final long id) {
                 String previous = mViewPagerAdapter.getSelectedPulseTarget();
-                mViewPagerAdapter.setSelectedPulseTarget(mSpinnerAdapter.getItem(position));
                 if (!previous.equals(mSpinnerAdapter.getItem(position))) {
-                    Timber.d("Switching chapter!");
-                    mViewPagerAdapter.notifyDataSetChanged();
+                    refreshSpinner(mSpinnerAdapter.getItem(position));
```
tasomaniac: ViewPager's notifyDataSetChanged is not working because it aggressively cache the Fragment state which shouldn't be cached because we are changing the list completely. Eg. first we show countries, then we show cities in a specific country.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
```diff
@@ -39,13 +40,18 @@
 
     private LayoutInflater mInflater;
     private ArrayList<Map.Entry<String, PulseEntry>> mPulse;
-    private int[] mPosition;
+    private int[] mPositions;
```
tasomaniac: When I disabled the ViewPager's aggressive caching, it solved the numbering problem for changing countries/cities. But it didn't solve when orientation change occurs. When orientation change occur in a page, the list scroll position is kept but the computed Positions were not kept. We may solve the problem with changing the algorithm to compute positions but it requires more work. I've just added kept the positions computed.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseFragment.java
```diff
@@ -76,13 +77,21 @@ public void onAttach(final Activity activity) {
...
     public void onActivityCreated(Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
 
         mTarget = getArguments().getString(ARG_TARGET);
         mMode = getArguments().getInt(ARG_MODE);
 
-        mAdapter = new PulseAdapter(getActivity());
+        int[] positions = savedInstanceState != null
+                ? savedInstanceState.getIntArray(INSTANCE_STATE_POSITIONS) : null;
+        mAdapter = new PulseAdapter(getActivity(), positions);
```
tasomaniac: Here I am saving the calculated positions of the adapter.

tasomaniac: By meaning positions, I mean that rankings of the countries/cities. The ones related to the numbering problem in the screenshot in the original issue.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
tasomaniac: Sure. Makes sense.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
stefanhoth: Renaming this private field changes the caching behavior?

tasomaniac: No :) While I was doing it I also refactored it. mPosition is not suitable because it was not plural. We are holding multiple items in it.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
tasomaniac: The type of the objects is `Map.Entry<String, PulseEntry>` not just `PulseEntry`. Map.Entry does not implement Comparable. So it will not work. Getting rid of `Map.Entry<String, PulseEntry>` is out of this PR's scope I think.

stefanhoth: This adapter manages an `ArrayList` of `Map.Entry`. WAT? :wink:

I'll try to address this later because it bothers me physically. :+1:

tasomaniac: I said OMG when I first saw it. :) It is everywhere in the adapter. The key in the map entry is actually used as the title.

stefanhoth: Tracked as #487

stefanhoth:
Ok, let's :ship: it.

PR #482, opened 2015-09-04
480/Add intent extras for custom tabs

friedger:
When opening an external link we add now custom tabs information for color and title.

Before | After
---- | -----
 |

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
+++ b/app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
@@ -16,7 +16,6 @@
 
 package org.gdg.frisbee.android.utils;
 
-import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageInfo;
@@ -27,7 +26,6 @@ import android.net.Uri;
 import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.customtabs.CustomTabsIntent;
-import android.support.v4.app.ShareCompat;
 import android.util.DisplayMetrics;
 import android.util.Patterns;
 import android.util.TypedValue;
@@ -36,6 +34,12 @@ import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonDeserializer;
+
+import org.gdg.frisbee.android.R;
+import org.gdg.frisbee.android.api.deserializer.DateTimeDeserializer;
+import org.joda.time.DateTime;
+import org.joda.time.Period;
+import org.joda.time.format.DateTimeFormat;
 
 import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


tasomaniac:
Does this still ask the user to open the link in native Google+ app? I think opening these in Google+ app has much better experience.

app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
```diff
@@ -218,7 +219,12 @@ public static boolean canLaunch(Context context, final Intent viewUrlIntent) {
 
     @SuppressWarnings("deprecation")
     public static Intent createExternalIntent(Context context, Uri uri) {
-        Intent intent = new Intent(Intent.ACTION_VIEW, uri);
+        CustomTabsIntent customTabsIntent = new CustomTabsIntent.Builder()
+                .setToolbarColor(context.getResources().getColor(R.color.theme_primary))
```
tasomaniac: You may also use this to access real primaryColor of the theme. I am not really sure which is the best. We are using different colorPrimary for DevFest, StudyJams etc. 

        ThemeUtils.getThemeAttrColor(context, R.attr.colorPrimary);

app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
```diff
@@ -218,7 +219,12 @@ public static boolean canLaunch(Context context, final Intent viewUrlIntent) {
 
     @SuppressWarnings("deprecation")
     public static Intent createExternalIntent(Context context, Uri uri) {
-        Intent intent = new Intent(Intent.ACTION_VIEW, uri);
+        CustomTabsIntent customTabsIntent = new CustomTabsIntent.Builder()
+                .setToolbarColor(context.getResources().getColor(R.color.theme_primary))
+                .setShowTitle(true)
+                .build();
+        Intent intent = customTabsIntent.intent;
+        intent.setData(uri);
```
tasomaniac: Don't we have to call `launchUrl` function of customTabs to make it work?

friedger:
The intent is not changed, only extras added. Same experience as before for all apps besides chrome.

app/src/main/java/org/gdg/frisbee/android/utils/Utils.java
friedger: No, launchUrl is just a convenience method

tasomaniac:
It is nice that nothing is changing. It can open Google+ like before. I didn't know it is just Intent Extras. Service binding is only for accessing internals and warm up. 

I've added this flag `FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET` before to display Google+ on its own task. Because of this CustomTabs look like outside of our Activity. Recent screen looks like this:

device-2015-09-05-012721

You see Chrome logo is showing. :/

tasomaniac:
I've removed `FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET` flag previously added by me. I hope you don't mind. If you can briefly review my commit we can merge it.

tasomaniac:
@friedger What do you think of the change I've committed? If it is ok feel free to merge this, if not feel free to revert it. :smile:

PR #486, opened 2015-09-07
TaggedEventSeries events do not need to be split.

tasomaniac:
Only the events in the main page namely `GdgEventListFragment` should be split.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
@@ -2,6 +2,7 @@ package org.gdg.frisbee.android.eventseries;
 
 import android.os.Bundle;
 import android.support.design.widget.Snackbar;
+import android.util.Pair;
 
 import org.gdg.frisbee.android.Const;
 import org.gdg.frisbee.android.R;
@@ -14,6 +15,9 @@ import org.gdg.frisbee.android.view.ColoredSnackBar;
 import org.joda.time.DateTime;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 import retrofit.Callback;
 import retrofit.RetrofitError;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -96,6 +100,45 @@ private void onNotFound() {
...
+     * Helper method that will split events into upcoming / past
+     * and add to the adapter of this list fragment
+     *
+     * @param events
+     *      The events list to be added
+     */
+    private void splitEventsAndAddToAdapter(Collection<? extends SimpleEvent> events) {
+        Pair<List<SimpleEvent>, List<SimpleEvent>> pair = splitEventsList(events);
+        mAdapter.addAll(pair.first);
+        mAdapter.addAll(pair.second);
```
stefanhoth: I'm a bit confused. Is that splitting basically a way of sorting or are they used somewhere separately?

stefanhoth:
I'm obviously not too familiar with the code base so excuse me. But what are tagged events and why aren't they split?

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
tasomaniac: We actually have sorting in the tagged events pages like devfest 's. Splitting is in the main page where we split past and upcoming events of A chapter.

stefanhoth: But first it's split and that added to the same adapter without any distinction. I'm just wondering of the same could be achieved by a comparator.

tasomaniac: Yeah exactly. That's what I said in the original PR when this is first
implemented. Splitting means here that showing the upcoming items first. It
would be better to do like that but there are too many things to do for me.

stefanhoth:
OK, LGTM. Thanks for the explanations.

tasomaniac:
Thank you for reviewing it :)

PR #496, opened 2015-11-24
Fix/App invite

friedger:
This PR fixes a problem with not showing the app invite.

The app invite was lost when the user was transferred from the start activity to the main activity.

After:
device-2015-11-24-085733

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: android
 android:
   components:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: android
 android:
   components:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,3 +1,4 @@
+sudo: false
```
tasomaniac: This is nice but it makes the build fail in Travis. The failures may be because of this.

build.gradle
```diff
@@ -6,10 +6,10 @@ buildscript {
         maven { url 'https://maven.fabric.io/public' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.4.0-beta1'
+        classpath 'com.android.tools.build:gradle:1.5.0'
```
tasomaniac: Finally they are out of beta!

tasomaniac:
Thank you for looking into this. Can you try to make the build pass and then I think it can be merged.

friedger:
@tasomaniac build passed! :-)

PR #500, opened 2015-12-12
Feature/App index

friedger:
This PR adds app indexing for the main site and events.

The title of the event is used as title of the content.
bildschirmfoto 2015-12-12 um 01 39 16

Searching for the event title in Google app will show the GDG app as a result with direct link to the event:
screenshot_2015-12-12-01-53-20
screenshot_2015-12-12-01-52-43


app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -286,24 +304,72 @@ public void failure(RetrofitError error) {
...
+        mViewPagerAdapter = new ChapterFragmentPagerAdapter(
+                this,
+                getSupportFragmentManager(), selectedChapterId
+        );
         mViewPager.setAdapter(mViewPagerAdapter);
         mViewPager.setOffscreenPageLimit(3);
         updateSelectionfor(selectedChapterId);
 
-        mTabLayout.setupWithViewPager(mViewPager);
+        recordStartPageView();
```
friedger: Not sure whether this is against best practise as the user did not really chose to view this content, but always gets back to it, e.g. after viewing an event.

However, it is recommended to have the main url gdgroups.org also indexed. Not sure which part of the app would map bets...

tasomaniac:
Wow we hit #500.

app/build.gradle
```diff
@@ -230,7 +237,10 @@ dependencies {
     //Test
     testCompile 'junit:junit:4.12'
     androidTestCompile 'org.assertj:assertj-core:1.7.1'
-    androidTestCompile ('com.android.support.test.espresso:espresso-core:2.2') {
+    androidTestCompile ('com.android.support.test.espresso:espresso-core:2.2.1') {
+        exclude module: 'support-annotations'
```
tasomaniac: You can also put conflict resolution above and we can remove these 2 excludes. I did it for Novoda droidcon London booth project. What do you think?

app/src/androidTest/java/org/gdg/frisbee/android/activity/EventActivityTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static android.support.test.espresso.intent.Intents.intended;
+import static android.support.test.espresso.intent.Intents.intending;
+import static android.support.test.espresso.intent.matcher.IntentMatchers.*;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.not;
+
+@RunWith(AndroidJUnit4.class)
+public class EventActivityTest {
```
tasomaniac: Moar tests!!! Thank you for this.

app/build.gradle
friedger: Good idea

app/src/main/AndroidManifest.xml
```diff
@@ -152,9 +161,22 @@
         <category android:name="android.intent.category.DEFAULT"/>
         <category android:name="android.intent.category.BROWSABLE"/>
       </intent-filter>
+      <intent-filter>
+        <action android:name="android.intent.action.VIEW"/>
```
tasomaniac: As far as I know, ParseDeepLinkActivity is related to Google+ share. 
I think we should move this intent-filter to `EventActivity`. What do you think?

app/src/main/java/org/gdg/frisbee/android/api/model/EventFullDetails.java
```diff
@@ -1,11 +1,15 @@
...
-import org.joda.time.DateTime;
+import com.google.gson.annotations.SerializedName;
 
 import java.util.ArrayList;
 
+import org.joda.time.DateTime;
+
 public class EventFullDetails {
-    String id, about, chapter, eventUrl, location, title;
+    @SerializedName("_id")
```
tasomaniac: Good catch. Wow.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -286,24 +304,72 @@ public void failure(RetrofitError error) {
...
+        );
+    }
+
+    private ResultCallback<Status> appIndexApiCallback(final String label) {
+        return new ResultCallback<Status>() {
+            @Override
+            public void onResult(Status status) {
+                if (status.isSuccess()) {
+                    Timber.d(
+                            "App Indexing API: Recorded event "
```
tasomaniac: Can you use Timber's beautiful formating to avoid String concatanation here.

app/src/main/java/org/gdg/frisbee/android/event/EventActivity.java
```diff
@@ -53,16 +70,86 @@ public void onCreate(Bundle savedInstanceState) {
...
+            result.setResultCallback(appIndexApiCallback("end " + viewAction));
+        }
+    }
+
+    private ResultCallback<Status> appIndexApiCallback(final String label) {
+        return new ResultCallback<Status>() {
+            @Override
+            public void onResult(Status status) {
+                if (status.isSuccess()) {
+                    Timber.d(
```
tasomaniac: It looks like we have duplicate code here. I am thinking to extract these App Indexing related classes to a super class.

tasomaniac:
Does this work after you open the event page and make it indexed? Or when you make this event indexed, am I going to get the result result as well? How does it work?
I actually build your branch but we have a problem with hub I think, it doesn't work currently.

app/src/main/AndroidManifest.xml
friedger: ParseDeepLinkActivity is for parsing all deep links. EventActivity expects an extra with the id. I'd prefer to have one exported activity that handles all the incoming intents...

friedger:
@tasomaniac When you open the event activity for a particular event this event get indexed. Do it a few times. Maybe even use the deep link helper in Android Studio to open this one event. Then open the Google App and start typing the title of the event. You might see a suggestion with the event title and the GDG app icon (like in the screenshot in this PR). Click on it (this helps to raise the ranking of the suggestion).

I don't know the suggestion algorithm and sometimes an event was suggested, sometimes it wasn't.

The suggestions produced by Google app depends on the user...

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
friedger: @tasomaniac Do you have an opinion on the indexing of the main activity?

friedger:
@tasomaniac Comments addressed

app/src/main/AndroidManifest.xml
tasomaniac: But deep linking integration for Chapter links are maintained in `MainActivity`. We should move that to here too then.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
tasomaniac: I have thought about it when I saw your comment here. Thank you for reminding. 
I am actually not sure. 

When I think normal webpages and SEO, main url's are always indexed more. I think it makes sense.

app/src/main/java/org/gdg/frisbee/android/event/EventActivity.java
```diff
@@ -53,16 +65,65 @@ public void onCreate(Bundle savedInstanceState) {
...
+    }
+
     protected String getTrackedViewName() {
         return "Event/" + getResources().getStringArray(R.array.event_tabs)[getCurrentPage()] 
-                + "/" + getIntent().getStringExtra(Const.EXTRA_EVENT_ID);
+                + "/" + getEventIdFrom(getIntent());
+    }
+
+    private void recordStartPageView() {
+        App.getInstance().getGdgXHub().getEventDetail(
```
tasomaniac: I just realized that we are doing this network call in EventDetailFragment too. 
So for every event detail page, we are doing the same network request twice.

friedger: I thought I can get away with it for this PR. Here is my suggestion for handling the communication between activity and fragment: da27744b742a92295596397c2990d60393a0c687

tasomaniac:
@friedger Can you check my last commit? 
I tried your last commit and it had a problem. Everytime I open an `EventDetail` page, it was waiting indefinitely and showing an error. When I go and come back, it was loading it immediately since it has the data in the cache. 
Right now, I reverted the data loading part. Fragment loads the data and tells Activity that loading is finished.

tasomaniac:
And I also realized that `recordEndPageView` functions are not working. 

The reason is that we are disconnecting from GoogleApiClient in `onPause` method. Since we are disconnecting, we just cannot send the data. We should look into this. What do you think?

tasomaniac:
I've just merged `develop` onto this to prevent the conflicts. 

`develop` branch has merged into this for 4 times. I think we should merge this. 

@friedger can you review 1 more time and merge this?

friedger:
@tasomaniac recordEndPageView works as the appIndex api is called before disconnection in `onStop`

tasomaniac:
@friedger Oh, I completely forgot that. It is supposed to work but I think it works async. I tracked logs and it sometimes logs saying that the indexing it successful. Most of the time, it logs nothing. 
For `recordStartPageView`, it logs every time without a miss.

friedger:
@tasomaniac I have tried with the event page and the stop event of main activity is logged after the event start event. This looks like it is cached and later synchronized.

Issue #355, opened 2015-05-08
Devices without GMS


tasomaniac:
For devices without GMS we are showing the following Dialog saying that "App won't work without Play Services installed". App runs fine without GMS excepts for the "Arrow" and "Achivements"

- [ ] We may display that Dialog only 1 time. Showing it on every page is really bad. 
- [x] Sign-in button in Settings should be disabled with a message about GMS is not available. 

![screen shot 2015-05-08 at 22 11 07](https://cloud.githubusercontent.com/assets/763339/7543764/bd52a8c8-f5cf-11e4-9646-0b05ed9ed43c.png)

friedger:
Related to https://github.com/gdg-x/frisbee/tree/fix/google_api

tasomaniac:
I believe the 2nd portion is done. I am not sure about the first though. Why are we showing that "Get Play Services" dialog? That is dumb. When there is no play services on the device, it is unlikely that the user could install it. Let's remove that dialog completely? What do you say?

friedger:
1st portion is not yet done.

How would you decide what to do on resolution error? Only check for SERVICE_MISSING? Started on https://github.com/gdg-x/frisbee/tree/355/no_gms

friedger:
1st portion can only be fixed when #483 is fixed as the Plus client shows the mentioned dialog automatically.

tasomaniac:
Are you sure that it shows it automatically? It gives the dialog to us and we show it using `showDialog` method.

friedger:
Not after my changes

tasomaniac:
Your changes already fixes the point 1 in this issue actually :)

friedger:
But the dialog still appears

tasomaniac:
It will make a huge difference if it only appear once. I think we can ignore that for now and we can open another issue for that if we want to remove it completely.

tasomaniac:
Fixed by #582

PR #501, opened 2015-12-12
Fix/Google api experience

friedger:
This PR improves the UX when Google Play services are not available
- the warning dialog about missing services is only shown once.
- the games menu entries are hidden when services are not available.
- the login settings is hidden when services are not available.

This PR also only asks for scopes/login when set in the preferences


friedger:
Fixes #355

Splaktar:
LGTM!

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
```diff
@@ -114,14 +114,16 @@ public void onCreate(Bundle savedInstanceState) {
...
-                .addApi(Games.API)
-                .addApi(AppStateManager.API)
-                .addScope(Plus.SCOPE_PLUS_LOGIN)
-                .addScope(Plus.SCOPE_PLUS_PROFILE)
-                .addScope(Games.SCOPE_GAMES)
-                .build();
+        GoogleApiClient.Builder builder = new GoogleApiClient.Builder(this)
+                .addApiIfAvailable(AppStateManager.API);
+        if (PrefUtils.isSignedIn(this)) {
+            builder.addApiIfAvailable(Plus.API, Plus.SCOPE_PLUS_LOGIN, Plus.SCOPE_PLUS_PROFILE)
```
tasomaniac: Why do we need this? I don't understand

friedger: With the scopes the play service asks for a user account, without scopes play services do not ask for login but we can still use the APIs for the PlusPersonDownloader and the news feed.

tasomaniac: For News feed actually, we cannot unfortunately use GoogleApiClient. It's because it is really limited. We have another dependency for Google plus. 

I am not sure but this comes to my mind. You are not adding those scopes for non logged in users. What happens when they click login? How do you add these scopes later?

friedger: The client is built when creating the GDG activity.

tasomaniac: I think this part makes problems. I am getting this crash:

```
java.lang.IllegalStateException: GoogleApiClient has an optional Plus.API and is not connected to Plus. Use GoogleApiClient.hasConnectedApi(Plus.API) to guard this call.
                                                                                   at com.google.android.gms.plus.Plus.zzf(Unknown Source)
                                                                                   at com.google.android.gms.internal.zzpr.getCurrentPerson(Unknown Source)
                                                                                   at org.gdg.frisbee.android.common.GdgNavDrawerActivity.updateUserPicture(GdgNavDrawerActivity.java:350)
```

friedger: UpdateUserProfilePicture should be updated, shouldn't it?

What should happen if play services are not available? The 'if applicable'
was added to handle this case better. Not connecting at all would result in
poorer experience for signed out users.

tasomaniac: I don't think we had a problem with signed out users. Signed out users can see all the data and the pictures from Google+. It's because we are using Google+ java api to get them.

tasomaniac: Btw, this crash happens after I login.

friedger:
@stefanhoth Do you have an opinion on this PR? 

We want to make the use of Google API more robust, make it work on emulators and devices without google services.

Is this the right approach to use `GoogleApiClient.Builder.addApiIfAvailable` and then guard the call to the api as in https://github.com/gdg-x/frisbee/commit/d1be9e64c5a64db909f652a5899dc997633902a0? 

Which API calls need to be guarded so that this does not happen https://github.com/gdg-x/frisbee/pull/501#discussion_r47772850

How to implement the logout?

tasomaniac:
There is new identity toolkit in GMS. AFAIK, it doesn't use GoogleApiClient at all for login. I am not sure how it goes with Google Play Games. When we migrate to that most of your questions will be answered. 

The new API does not use GoogleApiClient. 
It doesn't require for the user to have Google+

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -350,6 +355,9 @@ private void maybeRemoveUserPicture() {
     }
 
     private void updateUserPicture() {
+        if (!getGoogleApiClient().hasConnectedApi(Plus.API)) {
```
tasomaniac: I am not sure if this is the only place. How can we make sure that the other parts are ok too. I think we may have problems with other GMS calls we are doing.

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
```diff
@@ -130,10 +130,7 @@ protected void onStart() {
 
         mGoogleApiClient.registerConnectionCallbacks(this);
         mGoogleApiClient.registerConnectionFailedListener(this);
-
-        if (PrefUtils.isSignedIn(this)) {
-            mGoogleApiClient.connect();
-        }
+        mGoogleApiClient.connect();
```
tasomaniac: This introduced problems too. I've cleared data and setup the app without Signin. 
Since Google knows me and I have this app signed in elsewhere, because of this connect call, it automatically signed in. 
You may say that it is good but our app is not designed for that. The drawback is that `PrefUtils.isSignedIn()` stayed false eventhough I have signed in. The app thinks that I am not signed in.

friedger:
I don't think we want the identity toolkit, as we expect our users to have a google account, so moving to Google Sign-in would be sufficient: https://developers.google.com/identity/sign-in/android/

tasomaniac:
Yeah not the whole idendity tookit with everything. I meant the new Google Sign-in API in GMS 8.3.0. OMG we cannot use it since we are stuck at GMS 7.8 :(

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
friedger: Indeed, and how do I know that his "guard" is enough for the api to not complain about it. To much not-knowledge from my side.

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
friedger: We us the mGoogleApiClient to load chapter details e.g. in EventOverviewFragment. Hence, it feels wrong not to connect at all.

tasomaniac: You are right. And we are using `GoogleApiClient` for AppIndex API even when the user is not logged in. 

So then, what should we do? We can fix SharedPrefs thing with the login successful callback. But again, it will be weird, users will click "Skip Sign-in" and they will be signed in automatically since they recently signed in to the app.

friedger: Could it be improved by implementing a "proper" logout?

tasomaniac: I think we should implement "proper" logout and automatic sign-in. 

With automatic sign-in, we just need to update SharedPrefs and we should remove the login page from `FirstStartActivity` if the user automatically logins.

friedger:
I close this PR until we get a better understanding how to deal with login.
Here what I think:

We only need G+ login for the game.

Using login/logout of the GDG app should:
- connect/disconnect with Google Games
- reset the "Home chapter" to nearest chapter ignoring the value from the cloud

PR #506, opened 2015-12-16
Refactor progress bars

### LINKED ISSUE #466: Replace AnimationImageView with normal ProgressBar
We don't need `AnimationImageView`. It can be replaced with normal `ProgressBar` with custom `indeterminateDrawable` attribute. 

In fact I already started working on this.

tasomaniac:
This fixes #466 

I was working on a library to make ProgressBar's perfect. I've integrated that library. 
The library has a ProgressBar with show() and hide() methods. It show it after 500ms delay. When you call hide() in within that delay, it doesn't show it at all. This is really useful when the data comes from the cache. 
https://github.com/tasomaniac/DelayedProgress

I removed old `AnimationImageView` as well. 
The XML layouts were weirdly designed. They were using LinearLayout's `weight` feature to somehow automatically show/hide empty state, content and progress. 

I mostly made them explicit. You can see that after loading completes it shows the empty view automatically if the adapter is empty. 

Here is how it looks like: 
You can see that the Progress and content are shown/hidden with animation and it is shown only once. 

Event Series pages:
destination

Main page: 
destination


friedger:
LGTM

PR #507, opened 2015-12-17
Remove the profile picture if the user is not logged in anymore.

### LINKED ISSUE #406: on logout profile image in nav drawer should be removed.

tasomaniac:
Fixes #406


friedger:
The navdrawer activity shouldn't be responsible for this in this explicit way. Couldn't we rely on the user being null?

tasomaniac:
My fix is in NavDrawer Activity. 
The problem is that when the user logins, whe have `onConnected` callback. When the user logs out, there is no callback. So I am checking it in onResume.

friedger:
This does not work if the user revokes rights in the Google console, doesn't it?

tasomaniac:
Not sure. I don't think so. But I think that is out of this pr 's scope. We are saving the state in shared prefs. When user revokes it, we should delete our t data in shared prefs anyways.

friedger:
Lgtm

PR #508, opened 2015-12-17
Location sorting fix

### LINKED ISSUE #491: Chapters sorting when location is not available
When user's location is not available, the app does the same sorting with location of 0.0, 0.0 coordinates. 
Instead, don't take the location into account when it is not available. Just sort alphabetically.

tasomaniac:
Fixes #491 

Actually there were no problems. Location is null when there is no location. It is not 0,0. 
The problem was the chapter names starting with space char. I've added a deserializer to fix that. 

While doing that, 
Unnecessary comments are removed.
Location is given in constructor to the ChapterComparator. No need to use App class in the comparator.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/api/model/ChapterDeserializer.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/model/ChapterDeserializer.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013-2015 The GDG Frisbee Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gdg.frisbee.android.api.model;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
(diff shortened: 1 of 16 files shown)


friedger:
Not sure whether we should fix the chapters name, where do we stop fixing
issues? Preferably this is fixed in the hub!

tasomaniac:
Yes it should be done in hub. But Frisbee is in much better shape than hub. Hub has lots of issues. And I already implemented it :)

app/src/main/java/org/gdg/frisbee/android/api/model/ChapterDeserializer.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+
+import java.lang.reflect.Type;
+
+public class ChapterDeserializer implements JsonDeserializer<Chapter> {
+    @Override
+    public Chapter deserialize(JsonElement jsonElement, Type type, JsonDeserializationContext jsonDeserializationContext) throws JsonParseException {
+        Chapter chapter = new Gson().fromJson(jsonElement, type);
+        chapter.name = chapter.name.trim();
```
friedger: Too fast merged, it is about having two spaces between 'GDG'  and city name. Not about spaces that can be trimmed.

friedger:
Chapter.getShortname needs fixing

app/src/main/java/org/gdg/frisbee/android/api/model/ChapterDeserializer.java
tasomaniac: Do we have 'GDG'  keyword coming from the API for every chapter? In my manual tests, it seems that this PR fixes the problem described in the issue.

tasomaniac: I didn't quite follow. If you can describe the problem I can fix it :)

friedger: On Gitter

PR #509, opened 2015-12-18
Chapter sorting fix

tasomaniac:
Fixes mentioned in #508 are done


app/src/main/java/org/gdg/frisbee/android/api/model/Chapter.java
```diff
@@ -99,37 +99,42 @@ public String toString() {
     }
 
     public String getShortName() {
-        return name.replaceAll("GDG ", "");
+        if (shortName == null) {
+            shortName = name.replaceAll("GDG ", "").trim();
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/api/model/Chapter.java
```diff
@@ -99,37 +99,42 @@ public String toString() {
...
+        }
+        return shortName;
     }
 
     @Override
     public int compareTo(Chapter o) {
-        return name.compareTo(o.getName());
+        return getShortName().compareTo(o.getShortName());
```
friedger: should use the field instead of the getter

app/src/main/java/org/gdg/frisbee/android/chapter/ChapterComparator.java
```diff
@@ -32,7 +30,7 @@ public int compare(Chapter chapter, Chapter chapter2) {
         }
 
         if (lastLocation == null) {
-            return chapter.getName().compareTo(chapter2.getName());
+            return chapter.compareTo(chapter2);
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/api/model/Chapter.java
tasomaniac: It may be confusing but I had to do it like this. It is because the getter actually does the shortening and trimming when it is first called.

app/src/main/java/org/gdg/frisbee/android/chapter/ChapterComparator.java
friedger: Well, let's just use the default comparison! So nice!

friedger:
LGTM

friedger:
Maybe a test :-)

tasomaniac:
I am so bad. I am actually preparing for Android Testing codelab and I am submitting PRs without test. I promise I will write test for this.

PR #512, opened 2015-12-26
Tests are added for ChapterComparator.

tasomaniac:
Test coverage for ChapterComparator is %93 now. It is not %100 because of the wrong `distanceBetween` calculation made in Robolectric implementation.
Geo lat, lng getters are made public for testing.

BTW, targetSdkVersion is set to 21 because Robolectric does not have 22 yet. There is not much difference between 21 and 22 in terms of targetSdkVersion anyways.

### DIFF
--- a/app/src/test/java/org/gdg/frisbee/android/chapter/ChapterComparatorTest.java
+++ b/app/src/test/java/org/gdg/frisbee/android/chapter/ChapterComparatorTest.java
@@ -1,0 +1,117 @@
+package org.gdg.frisbee.android.chapter;
+
+import android.location.Location;
+import android.location.LocationManager;
+
+import org.gdg.frisbee.android.BuildConfig;
+import org.gdg.frisbee.android.api.model.Chapter;
+import org.gdg.frisbee.android.api.model.Geo;
+import org.gdg.frisbee.android.app.BaseApp;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricGradleTestRunner;
+import org.robolectric.annotation.Config;
+
+import static org.junit.Assert.assertTrue;
+
+@RunWith(RobolectricGradleTestRunner.class)
+@Config(constants = BuildConfig.class, application = BaseApp.class)
+public class ChapterComparatorTest {
+
+    private Location locationIstanbul;
+
+    private static final Chapter chapterWithoutLocation = new Chapter("GDG AAUA", "115737274430015604080");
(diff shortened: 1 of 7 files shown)


app/src/test/java/org/gdg/frisbee/android/chapter/ChapterComparatorTest.java
```diff
@@ -0,0 +1,117 @@
...
+        chapterIstanbul.setGeo(new Geo(28.97696, 41.00527));
+        chapterEskisehir.setGeo(new Geo(39.7766667, 30.5205556));
+    }
+
+    @Test
+    public void shouldBeAlphabeticalWhenNoHomeAndLocation() {
+        ChapterComparator comparator = new ChapterComparator(null, null);
+
+        //Chapter starting with digit should come before chapter starting with 'A'
+        assertTrue(0 > comparator.compare(chapterStartsWithDigit, chapterStartsWithLetterA));
```
friedger: I find this difficult to read, could you extract a method for these asserts? Like assertOrder(comparator, chapterStartsWithDigit, chapterStartsWithLetterA, chapterInstanbul);

tasomaniac: Definitely, thank you for the suggestion. It was in my mind but got list at the end.

tasomaniac:
How about now :) It really like this way. @friedger

app/src/test/java/org/gdg/frisbee/android/chapter/ChapterComparatorTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertElementsOrderedLikeThis(comparator,
+                chapterIstanbul, chapterWithDoubleSpaceInName);
+    }
+
+    private <T> void assertElementsOrderedLikeThis(
+            Comparator<? super T> comparator, T... elements) {
+        List<T> expectedOrder = Arrays.asList(elements);
+
+        List<T> shuffledAndSorted = new ArrayList<>(expectedOrder);
+        Collections.shuffle(shuffledAndSorted, new Random(0));
```
friedger: I don't like the Random here, as we only use two and three parametes we could have deterministic shuffling. However, LGTM.

tasomaniac: I got the code from @swankjesse and didn't change it. Here is the link. 
https://publicobject.com/2007/04/helper-method-for-unit-testing.html

But you are right. For our case, it doesn't add much and it may even cause problems.

PR #515, opened 2016-01-04
Use grgit gradle plugin for gitsha

friedger:
This PR removes dependency from particular git installation and makes it easier to work on Windows


friedger:
fixes #515

tasomaniac:
We are actually not using that `gitSha` property. We should have it somewhere in the About screen (maybe only on dev builds)

PR #521, opened 2016-01-08
All tests are green.

tasomaniac:
Link without id does not open MainActivity in our code. Can be reverted when that functionality implemented.
MainActivity test tries to switch to Istanbul. It fails when the default chapter is not set. I set to Brussels in the beginning.


app/src/androidTest/java/org/gdg/frisbee/android/activity/IntentFilterTest.java
```diff
@@ -41,15 +41,6 @@ public void mainActivityDoesNotLaunchWithInvalidUri() throws PackageManager.Name
     }
 
     @Test
-    public void gdgroupEventsLinkWithoutIdLaunchesMainActivity() {
```
friedger: We want this test. Should be DeeplinkActivity

friedger:
Deleted test should be re-added. When fixed LGTM.

tasomaniac:
Ok I will revert that in a bit.

PR #527, opened 2016-01-11
5137/Center events lists

friedger:
This PR centers the event lists

Before|After
---|---
layout-2016-01-11-115538|layout-2016-01-11-115509


tasomaniac:
Wow :) that easy :)

PR #529, opened 2016-01-13
Multidex implementation was wrong. `prodDebug` was not working at all.

tasomaniac:
I enabled multidex for flavour `dev` before. It was wrong. It should be enabled for `debug` build.
The difference between `dev` and `prod` is only the minApiLevel.


tasomaniac:
It seems that nobody used Kitkat or below emulator/device to develop something :)

friedger:
For me it said that my device does not fullfil minSdk

tasomaniac:
When you are on `devDebug` it only accept Lollipop and above. 
When you switch to `prodDebug` you can install on Kitkat devices now.

Issue #373, opened 2015-07-11
Improvement: GDG chapter page selection


PareshMayani:
There are thousands of GDGs and user has to scroll down to select the particular chapter, it's really not a good design I would say! 

Needs to make an improvement to make a selection easier! Either we can allow him to input text and then search!

![device-2015-07-11-153020](https://cloud.githubusercontent.com/assets/888080/8633376/4c0c2e38-27e2-11e5-94f5-97bdec7445c5.png)

tasomaniac:
Nice issue. I remember this is talked but it couldn't get into issue list. 
I think we should definitely do this. But it requires a little design change. How do you think we should do it?

PareshMayani:
@tasomaniac let me focus more on integrating design support library, I am midway integrating NavigationView, but meanwhile you can think :) will sure share mock-up later!

tasomaniac:
Oh nice, @HenriqueRocha also worked on integration of new NavigationView. He said he had problems with dynamic addition of the event series pages like DevFest, IO Extended

friedger:
For the chapter selection there is an ordering by distance.

The initial selection for the home chapter is done also by distance.

I think it is quite good as it is now.
Am 11.07.2015 14:32 schrieb "Said Tahsin Dane" <notifications@github.com>:

friedger:
If we have not access to the location we could show a autocomplete view
But this is really an edge case.
Am 11.07.2015 15:22 schrieb "Friedger Müffke" <friedger@gmail.com>:

PareshMayani:
@tasomaniac I am already done with navigation view with adding items dynamically! Sorry I didn't checked and know that he has tried implemented! But I have to make changes to proceed further, one issue I got soon after adding design support library, which is scrim layout and attributes, so I did change and able to proceed further!

tasomaniac:
I think scrim insets layout is not needed anymore with the new
NavigationView. Feel free to delete it.

PareshMayani:
@tasomaniac yeah that's what I did it and had to do to proceed further. I just made a commit yesterday for the NavigationView, do check and let me know if there is any issue!

friedger:
fixed by #505

tasomaniac:
Fixed by #505 and #530

PR #530, opened 2016-01-13
373/Add TextInputLayout for AutoCompleteSpinner for power users when typing chapter name

friedger:
This PR adds a hint when the chapter name has been entered partly but only one chapter name matches the entered text.
device-sr


tasomaniac:
TextInputLayout is good to show the hint while the user is typing. But I didn't understand why you added the chapter name into hint.

friedger:
fixes #373

friedger:
@tasomaniac You complained that the confirm button is enabled when the text is incomplete. The hint is now explaining why it is enabled.

tasomaniac:
:smile: I actually didn't understand what it is doing :) I now understand it. It is really for power users. It becomes enabled and selects that 1 option.

friedger:
the last commit was peer programmed with @tasomaniac

tasomaniac:
I will merge this when the checks are successful.

friedger:
Looks great!

PR #536, opened 2016-01-15
Fix 3 achievements

### LINKED ISSUE #533: New Achievement bugs
- [X] Achievement after sending a feedback does not work for me. 
- [x] When we migrate Feeling Social achievement, it worked for me but it unlocked level 3 but it skipped level 2. 
- [X] Power user is not implemented at all.

tasomaniac:
- KissesFromGdgXTeam
- FeelingSocial - 2 step achievements
- Power User

achievements are fixed. 

Fixes #533

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/appwidget/GdgDashClockExtension.java
+++ b/app/src/main/java/org/gdg/frisbee/android/appwidget/GdgDashClockExtension.java
@@ -22,6 +22,8 @@ public class GdgDashClockExtension extends DashClockExtension {
 
     @Override
     protected void onUpdateData(int i) {
+        PrefUtils.setWidgetAdded(this);
+
         final String homeGdg = PrefUtils.getHomeChapterId(GdgDashClockExtension.this);
 
         if (homeGdg == null) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/gdg/frisbee/android/appwidget/GdgDashClockExtension.java
```diff
@@ -1,64 +1,66 @@
...
+
+import java.util.ArrayList;
+
+import timber.log.Timber;
+
+public class GdgDashClockExtension extends DashClockExtension {
+
+    @Override
+    protected void onUpdateData(int i) {
+        PrefUtils.setWidgetAdded(this);
```
tasomaniac: I am not sure what caused this huge change but I only added this line.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -1,177 +1,177 @@
...
+        }
+
+        private void showChild(RemoteViews views, int i) {
+            views.setDisplayedChild(R.id.viewFlipper, i);
+        }
+    }
+
+    @Override
+    public void onUpdate(Context context, final AppWidgetManager appWidgetManager, int[] appWidgetIds) {
+        PrefUtils.setWidgetAdded(context);
```
tasomaniac: Here as well. Only this line.

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
```diff
@@ -103,6 +105,14 @@ public void onCreate(Bundle savedInstanceState) {
         addProperty("loggedIn", PrefUtils.isSignedIn(getActivity())); // Optionally add some properties
         addProperty("appStarts", PrefUtils.getAppStarts(getActivity()));
         buildProperties();
+
+        setRetainInstance(true);
```
friedger: Why do you want to retain the instance?

app/src/main/java/org/gdg/frisbee/android/appwidget/GdgDashClockExtension.java
friedger: It is the indenting. We should run reformat on the whole project.

app/src/main/java/org/gdg/frisbee/android/widget/FeedbackFragment.java
friedger: Better reset the achievement handler on detachActivity.

tasomaniac: I kinda had to. The problem is with the implementation of Doorbell's SDK we are using. 
They are dismissing the Dialog once the Send button is clicked and then they make the API request. In their Rest API class, they have a reference to Activity. 

So the problem is that achievement handler's method is run after `detachActivity` because of this. 

I wanted to change it completely. I wanted to make the dialog dismiss after API request finishes. Their callback only have `success` in it. We won't be notified if there is a network issue and the user will be stuck. 

I unfortunately couldn't figure out a better way because of this.

friedger: What about leaving the dialog open until doorbell comes back with a result?

There is also the `setErrorCallback`

tasomaniac: I didn't know that there is error Callback method. We can do that. It would
make sense. Btw the dialog does not survive orientation change.

friedger:
LGTM, waiting for the checks to compelete

PR #538, opened 2016-01-19
NullPointer fix. 

tasomaniac:
entries.length gave me NullPointerException since the hub is down now.

There is no need to save the size of the array. writeStringArray already saves the size and handle null state as well.


tasomaniac:
Here is the crash
https://fabric.io/gdgx/android/apps/org.gdg.frisbee.android/issues/569e2a57f5d3a7f76b2f9b08

PR #539, opened 2016-01-20
Continuous Distribution

### LINKED ISSUE #417: Setup continous distribution
local.properties and the p12 file is uploaded to travis.
It should be possible to publish apks by add the gradle task
```
publishApkRelease
```

tasomaniac:
It is finally working. I will write a blog post and a wiki about this. The setup is done now but any *maintainer* should be able to do this process from scratch. 

Here is what I do:

- Skip PRs and branches other than `master`
- Skip If the commit has a tag
- Checkout specified branch one again. 
- Increment version number (Used `perl` regex replace. Can be simpler)
- Store the new version as Major.Minor.Patch.Build (eg. 2.3.0.4)
- Create the ALPHA Apk again and publish to Play Store. 
- If successful, create a commit and push quitely with a tag. 

Fixes #417


tasomaniac:
The account that makes the commit is `support@gdgx.io`. It can be changed if requested.

tasomaniac:
And lastly, I thought about distribution for `develop` branch. That can be confusing with this setup. To tag every release we have, I am pushing a commit after a successful release. When we merge multiple PRs (which is quite common) Travis will build all of them in sync. It will create problems. 

- We can limit the async build limit in Travis to 1. Which will solve the issue, but it will make builds slower. 
- Or we can use 'master' branch or a special 'release' branch to do this.

app/build.gradle
```diff
@@ -81,7 +81,6 @@ android {
         }
         alpha.initWith(release)
         alpha {
-            debuggable true
```
friedger: Why did you remove debuggable?
Could be useful for alpha testers.

friedger:
So the process is merge develop into master and push it to master, correct?

friedger:
I don't think we need distribution of the develop branch

We should have a release branch that pushes to the beta channel and is manually pushed to prod.

app/build.gradle
tasomaniac: Play publisher does not create task and does not work if the type is debuggable :(

tasomaniac:
I've added methods to make it cleaner. 

We can adjust the script to set branch and track to deploy. 

master or release branches can deploy to alpha or beta. It's up to us. Let's decide.

PR #543, opened 2016-01-22
chapter event list is now fetched from hub.

### LINKED ISSUE #542: Chapter's events shouldn't use developers.google.com
Instead, it should use our hub.

tasomaniac:
Previously, i don't know why, we were using some hidden api from developers.google.com. I have been thinking to migrate it to hub. I finally did it.

What was the diff between `Event` and `TaggedEvent` classes? It's just that the former is coming from developers.google.com, the latter is from hub. Which does not make sense at all. Removed `TaggedEvent` class now. We only have `Event` class.

Fixes #542


app/src/main/java/org/gdg/frisbee/android/api/GroupDirectory.java
```diff
@@ -1,41 +1,33 @@
...
-public interface GroupDirectory {
-
-    @GET("groups/pulse_stats/")
-    Call<Pulse> getPulse();
-
-    @GET("groups/pulse_stats/{country}/")
-    Call<Pulse> getCountryPulse(@Path("country") String country);
-
-    @GET("events/feed/json")
-    Call<ArrayList<Event>> getChapterEventList(@Query("start") final int start,
```
tasomaniac: Only this endpoint is removed here.

friedger:
TaggedEvents are events that are tagged :-) 
There could be events that are not associated with a GDG in the list of tagged events. At least that is what I remember, like study jam or IO extended. For study jam, they have now removed this option.

tasomaniac:
They are just events at the end. If they are associated just their chapter
id will be null.

Splaktar:
:+1:

Issue #541, opened 2016-01-21
Events empty state alignment problem 


tasomaniac:
Here is the screenshot describing it. 
![screenshot_20160121-185321](https://cloud.githubusercontent.com/assets/763339/12487620/95c1039c-c070-11e5-8517-268a75743a5a.png)

friedger:
Assigned to @rciovati

tasomaniac:
Fixed by #549

Issue #546, opened 2016-01-25
Fix build on Travis for forked projects. 


tasomaniac:
Here are the details. Forked projects are failing because of gpg decryption problems. 
https://travis-ci.org/rciovati/frisbee/jobs/104715488

tasomaniac:
Fixed in #548

PR #549, opened 2016-01-26
fix 541: center the empy state image and text

rciovati:
This commit fixes the issue #541

### DIFF
--- a/app/src/main/res/layout/fragment_events.xml
+++ b/app/src/main/res/layout/fragment_events.xml
@@ -38,29 +38,14 @@
 
   <include layout="@layout/loading_animated_view"/>
 
-
-  <LinearLayout
+  <TextView
     android:id="@+id/empty"
     android:layout_width="wrap_content"
     android:layout_height="wrap_content"
-    android:layout_centerInParent="true"
-    android:gravity="center_vertical|center_horizontal"
-    android:orientation="vertical"
-    android:visibility="gone">
-
-    <ImageView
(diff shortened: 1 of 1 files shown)


rciovati:
Build is errored for the same reason of #546

friedger:
You could remove the LinearLayout all together

friedger:
FYI, naming convention for PRs is number/description 

Nothing important though

rciovati:
Thanks @friedger, I misread the contribution guidelines.

tasomaniac:
I've cherry picked your fix for Travis. Can you rebase your branch? The build should be fixed after that.

app/src/main/res/layout/fragment_events.xml
```diff
@@ -41,26 +41,18 @@
 
   <LinearLayout
     android:id="@+id/empty"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
-    android:layout_centerInParent="true"
```
tasomaniac: centerInParent is I think needed because we are using this in both RelativeLayout and FrameLayout time to time.

tasomaniac:
Since we don't have Image and text separately anymore, we can get rid of `LinearLayout` as well. What @friedger said :)

tasomaniac:
And nevermind my last comment about centerInParent.

tasomaniac:
Thank you for the contribution. :+1: 
This is really simple and not necessary but we usually put screenshots in PRs that change the UI.

PR #550, opened 2016-01-26
Fix plus profile image related crashes. 

tasomaniac:
PlusUtils class is introduced. …
Deprecated getCurrentPerson method is only used here.
It checks apiClient if it is connected and has necessary null checks.

- removeUserPicture method is moved inside updateUserPicture
This can be done now because we always have GoogleApiClient connected.


friedger:
LGTM

PR #554, opened 2016-02-03
Chore/Reformat and re-arrange

friedger:
I ran reformat with optimize imports and re-arrange items on the root folder.

This should prevent long diffs when only one line is changed.


app/src/main/java/org/gdg/frisbee/android/about/ExtLibrariesFragment.java
```diff
@@ -30,8 +30,9 @@
 import java.io.IOException;
 import java.io.InputStream;
 
-import butterknife.ButterKnife;
-import butterknife.Bind;
+import butterknife.Butte
```
tasomaniac: I think something wrong here

tasomaniac:
I am really afraid to merge this :) It is too long that I cannot review. And github cannot show the whole diff. If you are sure about this @friedger let's merge it.

friedger:
@tasomaniac If you feel like it you can check it out and review it :-)
All I did is using the IDE methods.

LGTM and move one...

PR #558, opened 2016-02-15
Git commit SHA and build time is generated only when we clean the build.

### LINKED ISSUE #553: Remove build timestamp from BuildConfig
Use paperwork instead of BuildConfig for gitSha and timestamp

Paperwork is available at
https://github.com/zsoltk/paperwork

tasomaniac:
When we release the app from CI, it will always be clean. So release versions of the app should always show the right SHA and time.

Before we were not even using those.
Not they are being sent to Crashlytics and shown in About screen only if the app is debug or alpha.

Fixes #553

screenshot_20160215-165313


app/src/main/java/org/gdg/frisbee/android/about/AboutFragment.java
```diff
@@ -1,58 +1,57 @@
...
+        return v;
+    }
+
+    @Override
+    public void onViewCreated(View view, Bundle savedInstanceState) {
+        super.onViewCreated(view, savedInstanceState);
+
+        mVersion.setText(BuildConfig.VERSION_NAME);
+
+        if (BuildConfig.DEBUG || BuildConfig.BUILD_TYPE.equals("alpha")) {
```
tasomaniac: Only this if statement is added.

app/src/main/res/layout-sw800dp-land/fragment_about.xml
```diff
@@ -1,46 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
tasomaniac: This is exactly the same as sw600dp version.

Splaktar:
Copyright year needs updating, 2013-2016. Should it be 'The GDG-X Frisbee Project'?

tasomaniac:
We should do it automated. I will look into it and update the whole
project.

friedger:
Lgtm

PR #566, opened 2016-03-05
minSdkVersion is now taken from command line property.

### LINKED ISSUE #565: Change multiDex flavor setup in favor of command line properties
Here is a gist that explains it. 
https://gist.github.com/neworld/6af9154591b24404c01d

When we use Google Developers' suggested way to speed up multidex builds, we sacrifice lint checks for minSdkLevel. The suggested way in the gist is perfect.

tasomaniac:
The default value for minSdkLevel is always 16. When Travis is publishing Apk to Play Store or we do a release build by command line, it will be 16. 

It can be easily turned to 21/22/23 for faster debug builds by providing `minSdk` property. 

This property can be set using Android Studio compiler settings. 

screen shot 2016-03-05 at 17 31 27

More info can be found [here](http://artemzin.com/blog/minsdk-without-flavors/)

Fixes #565

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,14 @@ Read the [Development Guide](https://github.com/gdg-x/frisbee/wiki/Developer-Doc
 
 When sending pull requests please make sure to enable EditorConfig in Android Studio -> Settings -> Editor -> Code & Style -> EditorConfig.
 
+####Speeding up debug builds
+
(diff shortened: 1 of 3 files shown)


stefanhoth:
:100: wow, this is amazing!

PR #572, opened 2016-03-13
All GCM related things removed.

tasomaniac:
GCM is not working and we have really old GCM implementation in our app. 

If we want to enable it in the future, we can re-write it.


app/src/main/java/org/gdg/frisbee/android/fragment/SettingsFragment.java
```diff
@@ -74,7 +63,7 @@
         public boolean onPreferenceChange(Preference preference, Object o) {
             final String homeGdg = (String) o;
 
-            if (mGoogleApiClient.isConnected() && PrefUtils.isGcmEnabled(getActivity())) {
+            if (mGoogleApiClient.isConnected()) {
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/utils/PrefUtils.java
```diff
@@ -197,16 +161,6 @@ public static void setAchievementSteps(@NonNull final Context context,
             .apply();
     }
 
-    public static boolean isAppStateMigrationSuccessful(final Context context) {
```
friedger: not gcm related

friedger:
I think we can still reuse most of the code for the new implemenation. However, the get the methods count down, LGTM

tasomaniac:
We know the code is here. It's been unused for years. Thank you for merging
:)

PR #573, opened 2016-03-13
Fix release builds

tasomaniac:
This will fix proguard and multidex issues we are having recently. 

#572 removes GCM and this removes location play services dependency. Once they are removed, we will be at 57K out of 65K method limit. 

After these 2 are merged, we can publish an alpha release.


app/google-services.json
```diff
@@ -1 +1,165 @@
...
\ No newline at end of file
+{
```
tasomaniac: Seems huge but this is generated by Google.

app/src/main/res/values/analytics.xml
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?><!--
-  ~ Copyright 2013-2015 The GDG Frisbee Project
```
tasomaniac: These are coming from Google Api Gradle plugins automatically

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.0.0-beta6'
-        classpath 'com.google.gms:google-services:1.3.1'
+        classpath 'com.google.gms:google-services:2.0.0-beta6'
```
friedger: should be 2.1.0-alpha1 :-)

app/build.gradle
```diff
@@ -64,12 +64,10 @@ android {
             proguardFile '../settings/proguard/proguard-square-retrofit2.pro'
             proguardFile '../settings/proguard/proguard-butterknife.pro'
             proguardFile '../settings/proguard/proguard-crashlytics.pro'
-            proguardFile '../settings/proguard/proguard-joda-time.pro'
```
friedger: Why is this gone? We shouldn't use joda anyway.

tasomaniac: JodaTime now includes its proguard configuration automatically through Gradle

app/build.gradle
```diff
@@ -142,9 +140,7 @@ dependencies {
     debugCompile 'com.android.support:multidex:1.0.1'
```
friedger: We still need multidex support then?

build.gradle
tasomaniac: I didn't do it because some people may not prefer the preview release

app/build.gradle
tasomaniac: Multidex is only on debug because we don't use proguard there.

tasomaniac:
What is holding you @friedger  :smile: I really want to release an alpha build by pushing to master.

friedger:
:+1:

tasomaniac:
LGTM

tasomaniac:
Now it can be merged, apparently this should be done after the last commit.

PR #584, opened 2016-03-17
Fix/581 crash

tasomaniac:
This is a trial to fix the crash. I someone can reproduce the problem easily, they can try the output of this work. 

If this makes sense to you all, we can release another ALPHA update with this to check at least on our devices including @PareshMayani


friedger:
:+1:

PR #587, opened 2016-03-17
583/Use REST api to retrieve chapter image

friedger:
This PR adds a retrofit service for the Google plus REST API.

It is used to retrieve chapter images in the event details page.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/api/model/Image.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/model/Image.java
@@ -1,0 +1,22 @@
+package org.gdg.frisbee.android.api.model;
+
+public class Image {
+    private String url;
+    private boolean isDefault;
+
+    public String getUrl() {
+        return url;
+    }
+
+    public boolean isDefault() {
+        return isDefault;
+    }
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
+    public void setDefault(boolean aDefault) {
+        isDefault = aDefault;
+    }
+}
--- a/app/src/main/java/org/gdg/frisbee/android/api/model/Image.java
+++ b/app/src/main/java/org/gdg/frisbee/android/api/model/Image.java
@@ -11,12 +11,4 @@ public class Image {
     public boolean isDefault() {
         return isDefault;
     }
-
-    public void setUrl(String url) {
-        this.url = url;
-    }
-
-    public void setDefault(boolean aDefault) {
-        isDefault = aDefault;
-    }
 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/gdg/frisbee/android/api/model/Image.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public String getUrl() {
+        return url;
+    }
+
+    public boolean isDefault() {
+        return isDefault;
+    }
+
+    public void setUrl(String url) {
```
tasomaniac: I would be safer to delete these setters. I guess you are not using them.

app/src/main/java/org/gdg/frisbee/android/api/model/ImageInfo.java
```diff
@@ -0,0 +1,13 @@
+package org.gdg.frisbee.android.api.model;
+
+public class ImageInfo {
+    private Image image;
+
+    public Image getImage() {
+        return image;
+    }
+
+    public void setImage(Image image) {
```
tasomaniac: Here as well.

app/src/main/java/org/gdg/frisbee/android/event/EventOverviewFragment.java
```diff
@@ -220,39 +218,48 @@ private void updateGroupDetails(Chapter group) {
...
-
-                                }
-                            });
+        loadChapterImageInto(group.getGplusId(), mGroupLogo);
+
+        ((GdgActivity) getActivity()).setToolbarTitle(group.getShortName());
+    }
+
+    private void loadChapterImageInto(String gplusId, ImageView groupLogo) {
+        App.getInstance().getPlusApi().getImageInfo(gplusId).enqueue(new retrofit2.Callback<ImageInfo>() {
```
tasomaniac: You should use our `Callback`. If you use this, necessary things like idlingResources won't happen.

friedger: @tasomaniac In https://github.com/gdg-x/frisbee/blob/583/anonymous_chapter_image/app/src/main/java/org/gdg/frisbee/android/api/PlusApiFactory.java#L18 there is the idlingResources configuration. Is that enough?

friedger: Your are right, but the reason is Timber logging, not idlingResources

tasomaniac: That's where idlingResources are registered. And then when the request is finished, they need to be released. We are using our Callback everywhere. So, if we use this here, it will make problems later.

tasomaniac:
:+1:

PR #588, opened 2016-03-17
483/Replace plus person downloader

friedger:
This PR replaces the plus api calls through PlusClient by calls through the retrofit service PlusApi

based on #587


friedger:
closed until #587 is merged

friedger:
With #589 this can be closed

app/src/main/java/org/gdg/frisbee/android/api/PlusApi.java
```diff
@@ -0,0 +1,18 @@
...
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Path;
+
+public interface PlusApi {
+    @GET("people/{gplusId}?fields=image&key=" + BuildConfig.IP_SIMPLE_API_ACCESS_KEY)
+    Call<ImageInfo> getImageInfo(@Path("gplusId") String gplusId);
+
+    @GET("people/{gplusId}?fields=image,aboutMe,tagline,urls,url,cover,displayName&key="
```
tasomaniac: Now I realized this duplication, this key can be added through OkHttp interceptors.

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -140,7 +143,7 @@ public void onCreate() {
         // Only the interceptors will be different.
         // We shouldn't have the below interceptor in other instances.
         OkHttpClient.Builder picassoClient = mOkHttpClient.newBuilder();
-        picassoClient.addInterceptor(new PlusPersonDownloader(plusClient));
+        picassoClient.addInterceptor(new PlusImageResizer());
```
tasomaniac: I don't understand the name. Does it resize the images?

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -401,7 +400,7 @@ private void maybeUpdateChapterImage() {
                     @Override
                     public Person doInBackground(String... params) {
 
-                        return PlusPersonDownloader.getPersonSync(params[0]);
+                        return App.getInstance().getPlusApi().getPerson(params[0]);
```
tasomaniac: Since it was difficult to work with Plus Client, it was not easy to have sync and async calls at the same time. 

Now that we have retrofit, we can just use Call class. If we use that we can call the network async or sync. If we can get rid of this AsyncTask and have async retrofit call here, that would be great. That can also be a TODO if you don't want to do it. 

What do you think?

friedger:
Comments addressed in #589

PR #589, opened 2016-03-17
483/Remove PlusClient dependencies

friedger:
This PR removes the dependency to the Google Plus Client (json/java) and removes the dependencies.

Now the app is below 65k methods and is used without multidex.

This PR includes also the changes of #587


app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class PlusImageResizer implements Interceptor {
```
friedger: This is what is left from the PlusPersonDownloader

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
```diff
@@ -198,24 +194,9 @@ public void onPostExecute(String[] params, ActivityFeed activityFeed) {
...
-    private ActivityFeed getActivityFeedSync(String param) {
-        try {
-            Plus client = App.getInstance().getPlusClient();
-
-            Plus.Activities.List request = client.activities().list(param, "public");
-            request.setMaxResults(10L);
-            request.setFields(
-                "nextPageToken,"
-                    + "items(id,published,url,object/content,verb,"
-                    + "object/attachments,annotation,object(plusoners,replies,resharers))");
```
friedger: This string has been moved into PlusApi annotation

tasomaniac:
Oh man, this is huge. I will review this later tonight. This is great!

app/src/main/java/org/gdg/frisbee/android/api/GapiOkHttpRequest.java
```diff
@@ -1,101 +0,0 @@
...
-
-import com.google.api.client.http.LowLevelHttpRequest;
-import com.google.api.client.http.LowLevelHttpResponse;
-import com.google.api.client.util.Preconditions;
-
-import java.io.IOException;
-import java.io.OutputStream;
-import java.net.HttpURLConnection;
-
-public class GapiOkHttpRequest extends LowLevelHttpRequest {
```
tasomaniac: Finally! :smile:

app/src/main/java/org/gdg/frisbee/android/api/SynchronousCallAdapterFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+import retrofit2.Call;
+import retrofit2.CallAdapter;
+import retrofit2.Response;
+import retrofit2.Retrofit;
+import timber.log.Timber;
+
+public class SynchronousCallAdapterFactory extends CallAdapter.Factory {
```
tasomaniac: Do we really need this? I mean it is good but is it needed? 
WIth Call class, we can do sync calls really easily.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
tasomaniac: This gives size to the api but the intention of this is not to resize the image. It gets the image from Google+ URL.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private static final Pattern mPlusPattern
+        = Pattern.compile("http[s]?:\\/\\/plus\\..*google\\.com.*(\\+[a-zA-Z] +|[0-9]{21}).*");
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+
+        Matcher matcher = mPlusPattern.matcher(request.url().toString());
```
tasomaniac: You removed the ModelCache from here. I think it is really needed. This is called in a ListView to load images from Google+. So cache would be awesome here.

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
```diff
@@ -92,20 +88,20 @@ public void onPreExecute() {
...
                 .setOnBackgroundExecuteListener(
-                    new CommonAsyncTask.OnBackgroundExecuteListener<String, ActivityFeed>() {
+                    new CommonAsyncTask.OnBackgroundExecuteListener<String, Activities>() {
                         @Override
-                        public ActivityFeed doInBackground(String... params) {
-                            ActivityFeed feed =
-                                (ActivityFeed) App.getInstance().getModelCache().get(Const.CACHE_KEY_NEWS + params[0]);
+                        public Activities doInBackground(String... params) {
+                            Activities feed =
+                                (Activities) App.getInstance().getModelCache().get(Const.CACHE_KEY_NEWS + params[0]);
```
tasomaniac: Since the classes in the cache are now changed, we shouldn't do this. 
We can either change the keys here. That will give us new space in cache. Or we can delete the cache on app update.

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
tasomaniac: I think we lost `setMaxResults` here. I wonder what is the default value. The default value can be better for us

tasomaniac:
I thought this would be much much more difficult. I thought we would have to refactor whole NewsFeed parsing etc. 

This is really awesome. I just have couple of comments and that's all.

tasomaniac:
BTW, this PR also includes the changes in #588 
I think that why you closed it before. I will close it now

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
friedger: You are right. What about `PlusImageFetcher`?

friedger: Or `PlusImageUrlConverter`?

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
friedger: Yes, default is 20. For Android N we could limit this value for Data Saving. I created an issue for that https://github.com/gdg-x/frisbee/issues/590

app/src/main/java/org/gdg/frisbee/android/api/SynchronousCallAdapterFactory.java
friedger: This was just to keep the changes to a minimum. I agree with you. And we should refactor the calls. Also the use of `Callback` is not consistent.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
friedger: I hoped that the http calls are cached. Needs verification.

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
friedger: Agreed, using a new space in Const.

app/src/main/java/org/gdg/frisbee/android/api/PlusApi.java
```diff
@@ -10,4 +14,15 @@
...
+
+    @GET("people/{gplusId}?fields=image,aboutMe,tagline,urls,url,cover,displayName&key="
+        + BuildConfig.IP_SIMPLE_API_ACCESS_KEY)
+    @Nullable Person getPerson(@Path("gplusId") String gplusId);
+
+    @GET("people/{gplusId}/activities/public?fields="
+        + "nextPageToken,"
+        + "items(id,published,url,object/content,verb,"
+        + "object/attachments,annotation,object(plusoners,replies,resharers))"
+        + "&key=" + BuildConfig.IP_SIMPLE_API_ACCESS_KEY)
```
friedger: Using key here (instead of in an interceptor) makes it more explicit where the authentication takes place. I prefer this solution of a

```
public class ApiKeyAdder implements Interceptor {
    @Override
    public Response intercept(Chain chain) throws IOException {
        Request original = chain.request();

        HttpUrl newUrl = original.url().newBuilder()
            .addQueryParameter("key", BuildConfig.IP_SIMPLE_API_ACCESS_KEY).build();
        Request.Builder requestBuilder = original.newBuilder().url(newUrl);

        Request request = requestBuilder.build();
        return chain.proceed(request);
    }
}
```

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -401,7 +400,7 @@ private void maybeUpdateChapterImage() {
                     @Override
                     public Person doInBackground(String... params) {
 
-                        return PlusPersonDownloader.getPersonSync(params[0]);
+                        return App.getInstance().getPlusApi().getPerson(params[0]);
```
friedger: For this PR, let's keep the changes small and refactor all these AsyncTask in the next step.

app/src/main/java/org/gdg/frisbee/android/Const.java
```diff
@@ -91,8 +91,8 @@
     public static final String CACHE_KEY_PULSE_GLOBAL = "pulse_global";
     public static final String CACHE_KEY_GDE_LIST = "gde_list";
     public static final String CACHE_KEY_FRISBEE_CONTRIBUTORS = "frisbee_contributor_list";
-    public static final String CACHE_KEY_PERSON = "person_";
-    public static final String CACHE_KEY_NEWS = "news_";
+    public static final String CACHE_KEY_PERSON = "person2_";
+    public static final String CACHE_KEY_NEWS = "news2_";
```
friedger: new model classes -> new cache keys

app/src/main/java/org/gdg/frisbee/android/api/PlusApi.java
```diff
@@ -1,13 +1,23 @@
...
+import org.gdg.frisbee.android.api.model.plus.ImageInfo;
+import org.gdg.frisbee.android.api.model.plus.Person;
 
 import retrofit2.Call;
 import retrofit2.http.GET;
 import retrofit2.http.Path;
 
 public interface PlusApi {
-    @GET("people/{gplusId}?fields=image&key=" + BuildConfig.IP_SIMPLE_API_ACCESS_KEY)
+    @GET("people/{gplusId}?fields=image")
```
friedger: ApiKeyAdder interceptor is now used

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -205,17 +136,50 @@ public void onNotFound(String key) {
...
-            for (Person person : people) {
-                addOrganizerToUI(person);
+    private void updateUIOnlineFrom(Person person) {
+        if (getActivity() != null) {
+            updateChapterUIFrom(person);
+            addOrganizers(person, true);
+        }
+    }
+
+    private void addOrganizers(final Person cachedChapter, boolean offline) {
```
friedger: This removed a lot of duplicate code

friedger:
We are hitting now User Rate limits for the people api on g+ in the infofragment on chapters with more than 5 organizers. We should move to Plus.PeopleApi.load that allows to retrieve all organizers in one call. @tasomaniac Do you agree?

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
tasomaniac: Yeah actually it is not a downloader. Converter makes really sense.

app/src/main/java/org/gdg/frisbee/android/api/SynchronousCallAdapterFactory.java
tasomaniac: It won't add much difference. With Call class, you will just add `execute` method and it will do the same. That's whats you are already doing with `getImageInfo` actually.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageResizer.java
tasomaniac: I should definitely needs verification. If plus have http cache, which I hope they have, then this will be even better.

app/src/main/java/org/gdg/frisbee/android/api/PlusApi.java
tasomaniac: I really like this interceptor. API keys and Auth related things usually done in interceptor anyways. That's what Square does

friedger: It is added now.

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -143,21 +152,15 @@ private void updateUIOnlineFrom(Person person) {
         }
     }
 
-    private void addOrganizers(final Person cachedChapter, boolean offline) {
```
tasomaniac: This is weird. I am sure that there were problems before. But now if we change this, we should go over all the usages of `addOrganizers`

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -407,7 +407,7 @@ public void onNotFound(final String key) {
                             @Override
                             public void success(Person person) {
                                 if (person != null) {
-                                    App.getInstance().getModelCache().put(key, person);
```
tasomaniac: :+1: We should look into the usages of `put` and make all of them async. The ModelCache causes a lot of issues in strict mode.

app/src/main/java/org/gdg/frisbee/android/api/OkClientFactory.java
```diff
@@ -45,10 +45,13 @@ public void log(String message) {
         return okHttpClient.build();
     }
 
-    public static OkHttpClient okHttpClientWithIdlingResources(OkHttpClient client) {
-        return client.newBuilder()
-            .addInterceptor(OkClientFactory.provideIdlingResourcesInterceptor())
-            .build();
+    public static OkHttpClient okHttpClientWithIdlingResources(OkHttpClient client, Interceptor ... interceptors) {
```
tasomaniac: The name of this method is now weird. It adds both idlingResources and any Interceptor given. 
I think these should be done separately.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageUrlConverter.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class PlusImageUrlConverter implements Interceptor {
```
friedger: Do we use this interceptor at all?

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -124,28 +93,38 @@ public void onActivityCreated(Bundle savedInstanceState) {
 
         final String chapterPlusId = getArguments().getString(Const.EXTRA_PLUS_ID);
         if (Utils.isOnline(getActivity())) {
-            new Builder<>(String.class, Person.class)
-                .setOnBackgroundExecuteListener(new CommonAsyncTask.OnBackgroundExecuteListener<String, Person>() {
+            App.getInstance().getModelCache().getAsync(Const.CACHE_KEY_PERSON + chapterPlusId,
```
tasomaniac: This line now exists in the else statement as well [here](https://github.com/gdg-x/frisbee/pull/589/files#diff-42ae8acf20f2f4d28c1a2703a94e0c2bR129)

I don't think this `isOnline` check is helpful. Because of that we have a hell of indentation here. I think shouldn't have `getAsync` from model 2 times. 

What we can do is that, we can first try to get it online, if that is unsuccesful, we can try to get from the cache, if that fails, we can use `setIsloading(false)`

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -153,48 +132,9 @@ public void onPostExecute(String[] params, Person person) {
...
-                                        getView(),
-                                        getString(R.string.bogus_organizer, org),
-                                        Snackbar.LENGTH_SHORT
-                                    );
-                                    ColoredSnackBar.alert(snackbar).show();
-                                }
-                            }
-                        }
+                        final Person cachedChapter = (Person) item;
+                        updateChapterUIFrom(cachedChapter);
```
tasomaniac: Why don't you use `updateUIOnlineFrom` here?

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -263,38 +230,53 @@ private Spanned getAboutText(Person person) {
         return Html.fromHtml(aboutText);
     }
 
-    private void updateOrganizersOnline(final Person person) {
```
tasomaniac: This part is really really hard to review. :( It was a mess before :( But right now it is still complicated.

app/src/main/java/org/gdg/frisbee/android/common/BaseFragment.java
```diff
@@ -55,4 +55,16 @@ protected void showError(@StringRes final int errorStringRes) {
             }
         }
     }
+
+    protected void showError(String errorString) {
+        if (isContextValid()) {
```
tasomaniac: We now have some duplication here. The other show error now can use this.

app/src/main/java/org/gdg/frisbee/android/api/PlusImageUrlConverter.java
tasomaniac: I did this and I am not very proud. This fixed the memory leak in Experts page before. 

We were doing network operation in the Adapter view handling before. That was causing memory leaks. In GDE page, we don't have imageUrl of all the people, we need to get them one by one. We only have their Google+ url. This converts Google+ URL into an image URL. 

We will change that outdated GDE API anyways. And it seems that it stopped working. So we will eventually get rid of this.

tasomaniac: http://gde-map.appspot.com/gde/list

now returns linkedin urls for social accounts instead of Google+. That's definitely a bug but we cannot trust it anymore. I suggest that we can disable GDE feature for this release. We can remove this ugly interceptor completely. 

And we can re-write GDE API later.

friedger: Let's do that in a different PR

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
friedger: Let's refactor this in a new PR.

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
friedger: updateUIOnline tries to fetch missing organizers from google, this branch doesn't

app/src/main/java/org/gdg/frisbee/android/common/BaseFragment.java
friedger: duplicate code has been removed

app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -164,7 +151,7 @@ private void addOrganizers(final Person cachedChapter, boolean online) {
                     } catch (Exception ex) {
                         if (isAdded()) {
                             addUrlToUI(url);
-                            showError(getString(R.string.bogus_organizer, org));
+                            Timber.w(ex, "Could not parse organizer " + org);
```
tasomaniac: :+1:

app/src/main/java/org/gdg/frisbee/android/api/PlusImageUrlConverter.java
tasomaniac: Definitely :+1:

tasomaniac:
LGTM

PR #592, opened 2016-03-18
591/GDE list model changed

### LINKED ISSUE #591: GDE list is empty
```
Expected a string but was BEGIN_ARRAY at line 1 column 15 path $[0].product
```

friedger:
This PR changes the server model, the server returns GDE models with product as String list.

This fixes #591


app/src/main/java/org/gdg/frisbee/android/api/model/Gde.java
```diff
@@ -14,11 +17,12 @@ public Gde createFromParcel(Parcel in) {
...
         }
     };
-    private String product, name, address, email, socialUrl;
+    private String name, address, email, socialUrl;
+    private List<String> product = new ArrayList<>();
     private double lat, lng;
 
     private Gde(Parcel in) {
-        product = in.readString();
+        in.readStringList(product);
```
tasomaniac: Does this expect product ArrayList not to be null? 
You can use this instead http://developer.android.com/reference/android/os/Parcel.html#createStringArrayList()

app/src/main/java/org/gdg/frisbee/android/api/model/Gde.java
```diff
@@ -38,8 +42,8 @@ public void writeToParcel(Parcel parcel, int i) {
         parcel.writeString(socialUrl);
     }
 
-    public String getProduct() {
-        return product.trim();
+    public List<String> getProduct() {
```
tasomaniac: Name this products maybe?

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
```diff
@@ -114,7 +120,7 @@ public void run() {
                 mViewPager.setAdapter(mViewPagerAdapter);
                 mTabLayout.setupWithViewPager(mViewPager);
 
-                getAchievementActionHandler().handleLookingForExperts();
+                //getAchievementActionHandler().handleLookingForExperts();
```
tasomaniac: Why was this commented out? Was that a mistake?

tasomaniac:
Looks good but I have couple of comments. 

I have another question tho. I saw that in the new API, all the social links are linkedin. That will make profile pictures stop loading. Do you have any comment on that?

tasomaniac:
Oh, I've checked the API now, and I see that it is already fixed. There is only one person with linkedin social url.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
friedger: We need to check whether the client has the appropriate API....

friedger:
This PR is only about products

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
friedger: Not related to the PR, but with the line the app breaka

app/src/main/java/org/gdg/frisbee/android/api/model/Gde.java
friedger: I had weird messages with gradles incremental builds. So, I left it unchanged.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
tasomaniac: Ok, but this is a general problem then. We cannot update the app until we fix it everywhere. Commenting this out will make it worse. I've just opened an issue for this #595

tasomaniac:
Image related thing is not an issue anymore. :smile: fortunately :crystal_ball:

friedger:
Feel free to fix and merge it. I'm off for the next two days

tasomaniac:
Ok, I will do it and merge then. I forgot my charger and I have limited time on my laptop :smile:

tasomaniac:
LGTM

PR #593, opened 2016-03-19
Remove unnecessary proguard rule. 

tasomaniac:
According to this, this problem is fixed. I heard somewhere that this can be removed. When it is removed, proguard will save use more space.

https://code.google.com/p/android/issues/detail?id=78377

I've alse searched for this crash in Fabric. We don't have the crash mentioned in the bug report. So it will be safe to remove this.


friedger:
:+1:

tasomaniac:
@friedger it says that it is out-of-date but you can still merge it. You don't have to update branch.

friedger:
From mobile I get a server error

PR #594, opened 2016-03-19
Remove url connection dependency

### LINKED ISSUE #564: Should remove httpurlconnection depedency
"The Apache HTTP client and HttpURLConnection APIs are deprecated. They continue to work as they always have, but we're moving everything to the new OkHttp 3 API. The okhttp-apache and okhttp-urlconnection modules should be only be used to accelerate a transition to OkHttp's request/response API. These deprecated modules will be dropped in an upcoming OkHttp 3.x release."

https://github.com/square/okhttp/blob/master/CHANGELOG.md

tasomaniac:
Since we removed Plus Client, okhttp-urlconnection dependency is not needed anymore.

Fixes #564

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -151,7 +151,6 @@ dependencies {
 
     //Square
     compile "com.squareup.okhttp3:okhttp:$rootProject.okHttpVersion"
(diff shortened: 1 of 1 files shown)


friedger:
Yes! :+1

friedger:
:+1:

PR #600, opened 2016-03-20
Unused SearchActivity and GdgProvider removed.

tasomaniac:
This SearchAcitivity and its provider were not used inside the app. This SearchActivity was crashing, nobody was able to open it.

It could be opened by really really old Android devices with no GMS. In the past, you could include your app inside global search by doing this. New Google search removed this functionality in favour of AppIndexing.

Fixes this Crashlytics issue http://crashes.to/s/80402380a5d


friedger:
No, no, no...

As we are going already down the GMS-free route, we should keep it. App Indexing is Google vendor lock-in, quick box search is open Android ecosystem and still available and documented (http://developer.android.com/guide/topics/search/search-dialog.html). 

You should use something like https://play.google.com/store/apps/details?id=com.startapp.quicksearchbox

friedger:
Chrome, Contacts, Google Play (named "Market - Market suggestions" w00t ??) exposes the search interface as well

tasomaniac:
Ok, you are right. But the Activity is crashing. It is not working at all. Let's fix it then. The fix is really easy actually.

PR #607, opened 2016-03-21
586/Make organizer check callback cancelable

### LINKED ISSUE #586: Crash when signing out
When signing out in settings:
http://crashes.to/s/0faa6a480a5

The settings activity was opened from the events page

friedger:
This PR integrates the organizer check callback better into the activity life cycle.

Fixes #586

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
@@ -101,6 +101,7 @@ public class MainActivity extends GdgNavDrawerActivity {
 
     // Local Broadcast receiver for receiving invites
     private BroadcastReceiver mDeepLinkReceiver = null;
+    private OrganizerCheckCallback organizerCheckCallback;
 
     /**
      * Called when the activity is first created.
@@ -197,6 +198,14 @@ public class MainActivity extends GdgNavDrawerActivity {
     }
 
     @Override
+    protected void onPause() {
+        if (organizerCheckCallback != null) {
+            organizerCheckCallback.cancel();
+        }
+        super.onPause();
+    }
+
+    @Override
     public void onConnected(Bundle bundle) {
         super.onConnected(bundle);
         updateChapterPages();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -205,21 +214,13 @@ public void onConnected(Bundle bundle) {
     private void updateChapterPages() {
         final boolean wasOrganizer = mViewPagerAdapter != null
             && mViewPagerAdapter.getCount() == ORGANIZER_PAGES.length;
+        if (organizerCheckCallback != null) {
```
tasomaniac: It would be great to initialize this at the beginning with `wasOrganizer = false` to make sure that it is never null.

friedger: It should be null whenever the task has finished.

tasomaniac: You never make it null again in the code.

friedger: Not sure how to do it from within the callback. It is set to null in onPause. (This does not prevent leaking the activity).

I prefer to think of the callback as a nullable, kind of "transient" object. I don't like to create an object to then throw it away (for the reason of clean code).

tasomaniac: That's the problem. onPause does not make it null.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -197,32 +199,36 @@ protected void onResume() {
...
-                @Override
-                public void onErrorResponse() {
-                }
-            }
+            organizerCheckCallback
         );
     }
 
+    private boolean isOrganizerFragmentShown() {
+        return mViewPagerAdapter != null
```
tasomaniac: :+1: this is good.

tasomaniac:
LGTM

PR #608, opened 2016-03-21
598/Google Sign In

### LINKED ISSUE #598: Signin does not work
In the first start it gives an error notification about play services. 
In settings, it does not do anything.

friedger:
This PR fixes the signin problem by inverting the result of `GoogleApiAvailability.getInstance().isUserResolvableError`
(I don't understand the meaning of this method name and documentation https://developers.google.com/android/reference/com/google/android/gms/common/GoogleApiAvailability.html#public-methods)

There are still some rough edges about the process but should work now.

Fixes #598

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
@@ -305,7 +305,7 @@ public abstract class GdgActivity extends TrackableActivity implements
     }
 
     private boolean isFatalPlayServiceError(int errorCode) {
-        return !GoogleApiAvailability.getInstance().isUserResolvableError(errorCode);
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
```diff
@@ -305,7 +305,7 @@ private void showFatalPlayServiceMessage(@NonNull ConnectionResult result) {
     }
 
     private boolean isFatalPlayServiceError(int errorCode) {
-        return !GoogleApiAvailability.getInstance().isUserResolvableError(errorCode);
```
tasomaniac: I don't understand this change. The previous version was more meaningful

friedger: But this works for me

PR #609, opened 2016-03-21
596/fix person cache

tasomaniac:
I was getting Unknown Organizer very quickly. After a quick investigation, I realized that the default expiration is 0. That means that when the user is online, cached content won't be used. 

I set expiration date for the people cached. If they changed their picture, they will be updated within a day. 

This actually solved Unknown Organizer problem. 

I also moved methods around to make it readable top down. This caused hard to read code changes. Sorry about that. :confused:

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
+++ b/app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
@@ -19,6 +19,7 @@ package org.gdg.frisbee.android.chapter;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.v4.content.ContextCompat;
 import android.text.Html;
 import android.text.Spanned;
 import android.text.SpannedString;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/gdg/frisbee/android/chapter/InfoFragment.java
```diff
@@ -291,11 +334,11 @@ private String getGPlusIdFromPersonUrl(Urls personUrl) {
                 .replace("u0", "")
                 .replace("https:", "")
                 .replace("http:", "")
-                .replace(plusId, "").replaceAll("[^\\d.]", "").substring(0, 21);
+                .replace(chapterPlusId, "").replaceAll("[^\\d.]", "").substring(0, 21);
         }
     }
 
-    private void setIsLoading(boolean isLoading) {
+    void setIsLoading(boolean isLoading) {
```
tasomaniac: This is removed because IDE suggested me to do it. I think I enabled a flag for this to be shown. Should we do these? What do you think? This will be better performance wise, but it can also make the code un-clean.

friedger: Yes, that is a nice change.

friedger:
:+1:

PR #610, opened 2016-03-21
598/fix login

### LINKED ISSUE #598: Signin does not work

tasomaniac:
It seems that `isUserResolvableError` has problem. I suggest to remove it completely.

When GMS is not available, `resolveIntent` will be null and the dialog will be shown anyways. There is no  difference in that case. 

This will cause conflicts with @friedger 's open PR #608 

Fixes #598


friedger:
If we don't trust the method we should implement it ourselves . The point was to only show the message once.

friedger:
If GMS is not available,  resilveIntent is not null and a dialog is shown to download play services.

tasomaniac:
For me when GMS is not available, I get a null intent with statusCode SERVICE_MISSING. It shows the error notification successfully. 

image

PR #612, opened 2016-03-22
Remove old code that shows old GMS error dialog.

### LINKED ISSUE #601: Play Services error on every app open
Play Services error unfortunately is shown in every app open. Before, it was shown in every page I think. Now it is only shown in MainActivity but it happens everytime I open the app. Here is a screenshot

screen shot 2016-03-21 at 11 22 36

tasomaniac:
The additional error dialog on every app open was caused by these lines in MainActivity. 

Fixes #601


friedger:
:+1:

PR #618, opened 2016-03-28
Fix crash related to TabLayout

tasomaniac:
This PR fixes Crashlytics crash 263
http://crashes.to/s/61ccaa3d9b9


friedger:
The chapterFragmentViewPagerAdapter contains the activity, so looks like still leaking

tasomaniac:
I don't think the crash here related to a leak. Maybe it is. I don't know. In either case this deprecated method should be removed I think. 
We can open an issue for the Adapter leaking.

PR #619, opened 2016-03-28
Remove model cache dir on app update

tasomaniac:
On app update the app now deleted the model cache directory. This will prevent crashes related to model changes in the future. We had a lot of these crashes in the past. 

Fixes crash 262
http://crashes.to/s/b5fd72e0102


friedger:
:+1:

friedger:
:+1:

PR #622, opened 2016-03-28
Fix crash related to DST

tasomaniac:
All dates are now UTC.

According to JodaTime, all date time objects should be valid. In some corner cases, because of the daylight saving times, some of them are invalid. When you create the object using UTC timezone, it will always be valid. This will fix crash 230, 264 http://crashes.to/s/3d775ef701f

More info
http://joda-time.sourceforge.net/apidocs/org/joda/time/IllegalInstantException.html
http://stackoverflow.com/questions/4983846/calculating-a-time-duration-with-joda-time


friedger:
:+1 :+1:

PR #623, opened 2016-03-28
Guard against NPE.

tasomaniac:
`isContextValid` is used to detect if the Fragment is still live after a network connection.

Fixes Crash 242
http://crashes.to/s/a08c9408cb3


friedger:
Why have you removed the visibility change on success? If it fails in a first attempt?

tasomaniac:
I don't think there is a point to make it INVISIBLE if it fails. Now it will just not load the image into ImageView which is very similat to making it INVISIBLE.

tasomaniac:
If you don't like it, I can revert it. :smile:

friedger:
The question is whether the fragment is reused. I don't know right now. When clicking through a tagged events list? Is it necessary to update the logo UI?

tasomaniac:
This is used in every event detail. Both when the user is coming from Tagged event lists, or from the main page.

We are putting chapter logo for some reason. That page should be re designed anyways.

friedger:
I meant reused without being recreated. Is it important that the UI is updated after it has been set at the first time.

friedger:
:+1:

PR #625, opened 2016-03-28
Several changes in PulseActivity. 

tasomaniac:
- selectedPulse cannot be null anymore. It is passed to adapter by the constructor.
- Its default value is GLOBAL
- When we retrieve a Pulse object, `setupPulseScreen` is called.
- `setupPulseScreen` sets up the array, inits the Spinner and selects the default selected value in Spinner.
- `initSpinner` now only inits the Spinner.
- `onItemSelected` checks if the previous selected item is available and then setup the ViewPager with `onPulseItemSelected`
- method ordering is adjust to make it top down readable.
- PulsePagerAdapter is now a static class to prevent Context leaks.

Supposed to fix crash 237
http://crashes.to/s/cdcf7ea59c7


friedger:
Still leaking IMHO. But LGTM

PR #630, opened 2016-05-01
Update libraries

tasomaniac:
ButterKnife is upgraded to 8.0.1 in a separate commit. 
ButterKnife has no longer an `unbind` method. To unbind, we now have to hold on to Unbinder object. Which is now done in BaseFragment automatically when the Fragments use `inflateView` method.

Build Tools, Fabric, Joda Time, Timber, Retrofit are also updated. They are only minor updates.


friedger:
Changes look good.
However, `wear:lint` failed.

friedger:
:+1:

PR #631, opened 2016-05-02
ButterKnife proguard rules are removed. 

tasomaniac:
ButterKnife proguard rules are not necessary anymore.
I forgot to remove them with the previous PR. Sorry :(


tasomaniac:
LGTM

Issue #628, opened 2016-04-22
Widget is not showing the next event


friedger:
Widget is not showing the next event (but the first one in the events list).

evindj:
I got this solved but I am having hard times to create the pull request. I keep getting a 403 error when I try to push the commit. Any help?

friedger:
Are you working on your own fork? Probably you do not have write
permission. You should fork on github and add the new repository as a new
remote url to your local git. Then push to that new remote repository. Then
create a pull request on github.com

Friedger

PR #639, opened 2016-05-19
widget is not showing the next event

evindj:


evindj:
Solves issue #628.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
                 });
 
         }
-
+        private Event getNextEvent(List<Event> eListEvents) {
+            if (eListEvents.size() == 0) {
+                return null;
```
friedger: If you return null the code below must handle this case as well

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
                 });
 
         }
-
+        private Event getNextEvent(List<Event> eListEvents) {
```
friedger: We try to have no prefixes for parameters

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -92,7 +92,7 @@ public void buildUpdate(final Context context, final AppWidgetManager manager, f
             final RemoteViews views = new RemoteViews(context.getPackageName(), R.layout.widget_upcoming_event);
             Intent mainIntent = new Intent(context, MainActivity.class);
             final PendingIntent pi = PendingIntent.getActivity(context, REQUEST_CODE_LAUNCH_FRISBEE,
-                mainIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+                    mainIntent, PendingIntent.FLAG_UPDATE_CURRENT);
```
tasomaniac: Can you please revert this line?

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
                 });
 
         }
-
```
tasomaniac: Can you please put a line break between methods?

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
...
+            }
+            Event ret = listEvents.get(0);
+            for (Event e : listEvents) {
+                if (e.getStart().isBefore(ret.getStart())) {
+                    ret = e;
+                }
+            }
+            return  ret;
+        }
         private void fetchEvents(Chapter homeGdg,
```
tasomaniac: Can you please put a line break between methods?

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
                 });
 
         }
-
+        private Event getNextEvent(List<Event> listEvents) {
```
tasomaniac: Can you annotate the return type with `@Nullable`.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
...
+            if (listEvents == null || listEvents.size() == 0) {
+                return null;
+            }
+            Event ret = listEvents.get(0);
```
tasomaniac: I didn't understand the for loop here. Can you explain a little bit? 
When you say next event I was waiting for the current day to be involved.

evindj: the list of events for the one month is requested from the backend
`.getChapterEventList(homeGdg.getGplusId(),
                    new DateTime(),
                    new DateTime().plusMonths(1))`
. Since I don't know the specifics of the implementations in the backend, I have to go through all the events in the return list to get the closest one in terms of start day and that will be the next event.

tasomaniac: We are looking for the closest future one to today. I am confused because you are not using today's date in this algorithm. 
I am pretty sure that the list is sorted once we got the list from the backend. The problem is that we separated the past and future events. When we did that separation, we didn't split the list into 2. We just changed the ordering. 
Maybe we really should need 2 separate lists. 

This method will be confusing to everybody in the future.

evindj: Maybe, I did not understood what the getChapterEvenList(DateTime start, DateTime finish) method of the backend does. my assumption was that it returns all the events between start and finish. new Datetime() is set to the current system millisecond in the default time zone(that is the equivalent of now I think). `App.getInstance().getGdgXHub() .getChapterEventList(homeGdg.getGplusId(), new DateTime(), new DateTime().plusMonths(1))` is therefore going to retrieve events between today and today + 1 month. From this list, it is either we loop through to get the closest date or if the list is sorted it will be the first element. I don't think it is necessary to separate the list in the backend. It will be very useful if we can confirm that the list from the backend is sorted and please do correct me if I have a wrong assumption on how the getChapterEventList works

tasomaniac: I didn't mean that we should separate them in the backend. Unfortunately, backend is really hard to touch. 

At some point last year, we separated the past and future events in the chapter events page. I don't know the exact specification of the backend but it returns past events as well. If you go to the main page in the app, you will see the past events. 

We basically sorted the list to seprate the past and future events. The issue described in #628 was a side effect of that separation. 

This method is not clear and feels like a patch for that side effect. Don't get me wrong, your code is fine. It's just that the codebase is not in perfect situation and we need help. 

And lastly, in this method, I still don't see the current date in the comparison. How do you make sure that the event returned here is the closest future event?

evindj: Great, it completely makes sense. I did the changes in the code accordingly.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -138,19 +161,23 @@ public void success(PagedList<Event> eventsPagedList) {
...
-                                firstEvent.getStart().toLocalDateTime()
-                                    .toString(DateTimeFormat.patternForStyle("MS",
-                                        getResources().getConfiguration().locale)));
-                            showChild(views, 1);
-
-                            Intent i = new Intent(UpdateService.this, EventActivity.class);
-                            i.putExtra(Const.EXTRA_EVENT_ID, firstEvent.getId());
-                            views.setOnClickPendingIntent(R.id.container,
-                                PendingIntent.getActivity(UpdateService.this, 0, i, 0));
+                            Event firstEvent = getNextEvent(events);
```
tasomaniac: Sorry for the late comment here, it is not that important but it would be great if we change the variable to `nextEvent`. It is not the first event anymore.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -124,6 +124,29 @@ public void onNotFound(String key) {
...
 
+        @Nullable
+        private Event getNextEvent(List<Event> listEvents) {
+            if (listEvents == null || listEvents.size() == 0) {
+                return null;
+            }
+            Event ret = listEvents.get(0);
+            for (Event e : listEvents) {
+                if (e.getStart().isAfterNow()) {
+                    if (ret.getStart().isBeforeNow()) {
```
tasomaniac: You are starting the `ret` value with the first value in the list. That's why it can be "before now" in all the other cases, it cannot before now, because there is a check in 134. 

I think this can be simplified. Checking "before now" does not make sense. 

Can you try to make it `null` first and handle the `null` in the for loop?

tasomaniac:
LGTM

tasomaniac:
I merged the PR but the real problem is with our backend. If we go into Berlin in the main app, the events list is not sorted there as well.

Issue #599, opened 2016-03-20
Fix inactive chapters click behavior in Pulse


tasomaniac:
Here is the problem. The app gets the Pulse page from DevSite. Even if the chapter is not active, it is available in Pulse. This means that there are more chapters in Pulse than the main page. 

Mexico City is a good example to reproduce it today. 

When you click Mexico City in Pulse, it opens your default main page. The reason is that it cannot find it. Instead of this, we should warn user when we couldn't find it. 

It happens for very few of them but this is very confusing to the users.

tasomaniac:
I've just looked at the devsite this morning, 

Mexico City is indeed listed as inactive. But the website can open and show it. 
https://developers.google.com/groups/chapter/111114679803991832261/

What should we do? Hub does not ingest inactive chapters.

friedger:
In Pulse, a snackbar could be shown that no information are available for this chapter.

friedger:
Inactive chapters are not defined by properties of the pulse object. Instead, here should be a check against the chapter directory whether the chapter is available locally. The hub only returns active chapters.

friedger:
fixed by #641

PR #640, opened 2016-06-03
599/bug

evindj:
Created the snackbar to display the message if the chapter is Inactive(has not organized any meetings in the last 6 months.)


app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -92,7 +92,7 @@ public void buildUpdate(final Context context, final AppWidgetManager manager, f
             final RemoteViews views = new RemoteViews(context.getPackageName(), R.layout.widget_upcoming_event);
             Intent mainIntent = new Intent(context, MainActivity.class);
             final PendingIntent pi = PendingIntent.getActivity(context, REQUEST_CODE_LAUNCH_FRISBEE,
-                mainIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+                    mainIntent, PendingIntent.FLAG_UPDATE_CURRENT);
```
tasomaniac: Hey I think your code style settings is different then the default. We are using the default settings and we have editor config support which should be applied automatically. 

You can change the style setting project by project. Can you choose the default settings for Frisbee. 

Please do not include changes you didn't make. Thank you :)

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
```diff
@@ -123,7 +123,18 @@ public void onNotFound(String key) {
                 });
 
         }
-
+        private Event getNextEvent(List<Event> eListEvents) {
```
tasomaniac: I am confused. Why does this show up as a new code?

tasomaniac:
@evindj thank you for the effort. 

It is a great PR but not what we want. We are not interested in chapters with 0 meetings. Chapters with 0 meetings can be active, right? 

The error checking should happen in `MainActivity`. 

When we open `MainActivity` with `Const.EXTRA_CHAPTER_ID`, it tries to find that chapter. When it cannot find it, it just shows the default one. Instead we want to show an error there. 

Errors can be 
- Dialog showing the error and pressing OK would close the Activity. 
- Just show a Toast and close the Activity.

tasomaniac:
And lastly, for the PRs, 

Can you make the titles more definitive? And possibly with more description with screenshots/gifs showing the work you done.

app/src/main/java/org/gdg/frisbee/android/appwidget/UpcomingEventWidgetProvider.java
friedger: I think because there are merge conflicts. @evindj could you use the current develop branch and add your changes on top?

friedger:
#599 is not very precise about what the solution would be. There should be a check against the chapter directory whether the chapter is available locally. The hub only returns active chapters.

I have added that to #599

tasomaniac:
Yeah, that's totally miss communication. Pulse does not know anything about active chapters. That's why the error case should be handled in MainActivity. 

@friedger What do you think about the error message implementation with a Dialog? I think that would be quite good for us.

friedger:
@tasomaniac The user should not be directed to the invalid screen and back again. Checking in Pulse should be better.

The dialog in MainActivity could be used for deep links.

tasomaniac:
It happens very few times. I think going to main activity and coming back is fine. 
Implementation wise it would be much simpler. Because otherwise Pulse screen would have to go to Hub to check it.

tasomaniac:
Oh you already did it. Cool.

friedger:
There are more inactive chapters than you think...

tasomaniac:
Fixed by #641

PR #641, opened 2016-06-03
599/Add check for inactive chapters in pulse

friedger:
This PR adds a check for inactive chapters.

If a chapter is inactive the city is shown as disabled and onClick shows a Snackbar instead of going to the chapter screen.

Screen showing shortly after clicked on GDG Mexico city:
device-2016-06-03-122621


friedger:
fixes #599

app/src/main/java/org/gdg/frisbee/android/pulse/PulseFragment.java
```diff
@@ -162,10 +164,23 @@ public void networkFailure(Throwable error) {
         }
     }
 
-    private void initAdapter(Pulse pulse) {
-        mAdapter.setPulse(mMode, pulse);
-        mAdapter.notifyDataSetChanged();
-        setIsLoading(false);
+    private void initAdapter(final Pulse pulse) {
```
tasomaniac: You are getting the chapter list after everything is ready. If we always need the chapter list, what about getting it upfront? 
First do the chapter retrieval and then do the rest. 
Then we can provide the Directory with the constructor of the adapter. Then we can make it private final. 

What do you think?

app/src/main/java/org/gdg/frisbee/android/pulse/PulseFragment.java
```diff
@@ -180,9 +195,15 @@ public void onListItemClick(ListView l, View v, int position, long id) {
         if (mTarget.equals(GLOBAL)) {
             mListener.openPulse(pulse.getKey());
         } else {
-            Intent chapterIntent = new Intent(getActivity(), MainActivity.class);
-            chapterIntent.putExtra(Const.EXTRA_CHAPTER_ID, pulse.getValue().getId());
-            startActivity(chapterIntent);
+            if (v.isEnabled()) {
```
tasomaniac: What about just disabling the click? It is grayed out. I think it would be simpler that way. 
We will need translation for the new string.

friedger: Ok, but the user might not understand why it is greyed out.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseFragment.java
friedger: makes sense.

friedger:
Issues addressed

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
```diff
@@ -38,16 +39,20 @@
 
 class PulseAdapter extends BaseAdapter {
 
-    private LayoutInflater mInflater;
-    private ArrayList<Map.Entry<String, PulseEntry>> mPulse;
-    private int[] mPositions;
+    private final LayoutInflater mInflater;
+    private final ArrayList<Map.Entry<String, PulseEntry>> mPulse;
+    @Nullable
```
tasomaniac: 👍  for Nullable.

app/src/main/java/org/gdg/frisbee/android/pulse/PulseAdapter.java
```diff
@@ -38,16 +39,20 @@
...
+    private final Directory mDirectory;
 
+    private int[] mPositions;
     private int mMode;
+    private boolean mCheckValuesAgainstDirectory;
 
-    public PulseAdapter(Context ctx, @Nullable int[] positions) {
+    public PulseAdapter(Context ctx, @Nullable int[] positions, Directory directory) {
```
tasomaniac: I think unfortunately the Nullable defined in field is not transitive to here. Should we add `@Nuallable` to the constructor param as well?

app/src/main/java/org/gdg/frisbee/android/pulse/PulseFragment.java
```diff
@@ -162,8 +178,12 @@ public void networkFailure(Throwable error) {
         }
     }
 
+    private boolean isGlobalSelected() {
```
tasomaniac: :+1:

tasomaniac:
LGTM

Splaktar:
What about just removing / hiding the entries for inactive chapters? They are most always 0 value anyway?

friedger:
values differ from 0 e.g. for circle counts. I would say that filtering is out of scope of the issue.

PR #642, opened 2016-06-05
Various refactorings

tasomaniac:
On the plane I've done some refactorings offline. :) 
I will explain them inline.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
@@ -96,9 +96,20 @@ public class AboutActivity extends GdgActivity {
                 // as the ID will be consistent on the sending and receiving devices.
                 String[] ids = AppInviteInvitation.getInvitationIds(resultCode, data);
                 Timber.d("Sent %d invitations", ids.length);
+                sendAnalyticsEvent(
+                    "AppInvite",
+                    "Successful",
+                    String.valueOf(ids.length),
+                    ids.length);
             } else {
                 // Sending failed or it was canceled, show failure message to the user
-//                showMessage(getString(R.string.send_failed));
+                showError(R.string.invitation_error_message);
+
+                sendAnalyticsEvent(
+                    "AppInvite",
+                    "Error",
+                    ""
+                );
             }
         }
     }
--- a/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
@@ -51,7 +51,7 @@ public class AboutActivity extends GdgActivity {
         getActionBarToolbar().setTitle(R.string.about);
         getActionBarToolbar().setNavigationIcon(R.drawable.ic_up);
 
-        mViewPager.setAdapter(new AboutPagerAdapter(this, getSupportFragmentManager()));
+        mViewPager.setAdapter(new AboutPagerAdapter(getSupportFragmentManager(), getResources()));
         mTabLayout.setupWithViewPager(mViewPager);
     }
 
@@ -86,7 +86,7 @@ public class AboutActivity extends GdgActivity {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
-        Timber.d("onActivityResult: requestCode=" + requestCode + ", resultCode=" + resultCode);
+        Timber.d("onActivityResult: requestCode= %d, resultCode= %d", requestCode, resultCode);
 
         if (requestCode == REQUEST_INVITE) {
             if (resultCode == RESULT_OK) {
--- a/app/src/main/java/org/gdg/frisbee/android/about/AboutPagerAdapter.java
+++ b/app/src/main/java/org/gdg/frisbee/android/about/AboutPagerAdapter.java
@@ -1,6 +1,6 @@
 package org.gdg.frisbee.android.about;
 
-import android.content.Context;
+import android.content.res.Resources;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
```diff
@@ -96,9 +96,20 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                 // as the ID will be consistent on the sending and receiving devices.
                 String[] ids = AppInviteInvitation.getInvitationIds(resultCode, data);
                 Timber.d("Sent %d invitations", ids.length);
+                sendAnalyticsEvent(
```
tasomaniac: Started sending analytics events about app invite.

app/src/main/res/values/strings.xml
```diff
@@ -184,6 +184,7 @@
     <string name="action_deep_link" translatable="false">
         org.gdg.frisbee.android.DEEP_LINK_ACTION
     </string>
+  <string name="invitation_error_message">App Invite was unsuccessful.</string>
```
tasomaniac: I wanted to enable the error in AppInvite flow. But this introduced another String. Do you think we can use one of the existing ones.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
```diff
@@ -94,102 +84,66 @@ public void networkFailure(Throwable error) {
...
-                mTabLayout.setupWithViewPager(mViewPager);
-
-                getAchievementActionHandler().handleLookingForExperts();
-            }
-        });
-    }
-
-    protected String getTrackedViewName() {
-        return null;
+    private void setupGdeViewPager(GdeList directory) {
```
tasomaniac: I've changed `addGdes` to `setupGdeViewPager`. Because here I do all the setup now. This way we are creating the adapter with a final immutable list.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
```diff
@@ -94,102 +84,66 @@ public void networkFailure(Throwable error) {
...
-        @Override
-        public Fragment getItem(int position) {
-            if (position == 0) {
-                return PlainLayoutFragment.newInstance(R.layout.fragment_gde_about);
-            } else {
-                String key = mGdeMap.keySet().toArray(new String[mGdeMap.size()])[position - 1];
-                Fragment frag = GdeListFragment.newInstance(mGdeMap.get(key));
-                mFragments.append(position, new WeakReference<>(frag));
+    private HashMap<String, GdeList> extractCategoriesFromGdeList(GdeList directory) {
+        HashMap<String, GdeList> gdeMap = new HashMap<>();
```
tasomaniac: Unfortunately the diff here looks bad. 
This method is basically the same. I just extracted this into a method. 
This gets the whole list and split it into categories.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
```diff
@@ -94,102 +84,66 @@ public void networkFailure(Throwable error) {
...
-                return "";
+                gdeMap.get(product).add(gde);
             }
         }
+        return gdeMap;
+    }
 
-        @Override
-        public void onPageScrolled(int i, float v, int i2) {
+    private List<GdeCategory> convertCategoryMapToList(HashMap<String, GdeList> gdeMap) {
```
tasomaniac: Using the HashMap in the List Adapter is not that great. Instead I am converting again the HashMap to a List of categories.

app/src/main/res/layout/fragment_get_involved.xml
```diff
@@ -54,6 +54,6 @@
     android:layout_height="wrap_content"
     android:layout_alignParentRight="true"
     android:layout_alignParentTop="true"
-    android:elevation="4dp"
```
tasomaniac: This fixes the "Fork me" badge in this page. Elevation in material design puts the card over this Image. Just a random 8dp elevation fixed the problem. I don't know what else we can do here if we don't like this. 
In devices below Lollipop, it is fine.

app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/about/AboutPagerAdapter.java
```diff
@@ -1,18 +1,19 @@
...
 class AboutPagerAdapter extends FragmentStatePagerAdapter {
-    private Context mContext;
 
-    public AboutPagerAdapter(Context ctx, FragmentManager fm) {
+    private final String[] pageTitles;
+
+    AboutPagerAdapter(FragmentManager fm, Resources resources) {
         super(fm);
-        mContext = ctx;
+        pageTitles = resources.getStringArray(R.array.about_tabs);
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
friedger: I am not a fan of this extra class shuffeling. 10+ extra object creations for nothing.

app/src/main/res/values/strings.xml
friedger: We should not be afraid of translations...

tasomaniac: :+1: on that. I just don't like semi translated apps. :)

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
tasomaniac: Using `HashMap` in the adapter is also really really bad. I was wondering about the ordering of the categories in the ViewPager titles. The ordering was random. I think that is also because of this HashMap. 

I can move this convertion code to a background thread.

friedger: The ordering is the ordering of the list, isn't it? We get the hash map
from the server. We could do some smarter conversions there. The adapter is
using a list.

tasomaniac: Ordering of the GDEs are also wrong but I was more talking about the
ordering of the categories.  It's also random.

The adapter had the HashMap before. And it was trying to access the key by
position which is really bad.

friedger:
:+1:

friedger:
Let's move on.

app/src/main/java/org/gdg/frisbee/android/gde/GdeActivity.java
friedger: Yes, you are right. I thought we had the same setup as in PulseAdapter.

Issue #442, opened 2015-08-01
Implement the new FCM registration


tasomaniac:
It can be related to this. 
https://plus.google.com/+AndroidDevelopers/posts/TqPcucE9zL9

We should check if it is the same case in 2.1 production version. 

Do we even have access to push notification server of ours? Is it the hub webpage?

Splaktar:
It looks like we've got about half of the migration done (receiver) but still need to update the other half (service).

The problem seems to be related to this:

    public class GcmIntentService extends IntentService

From https://developers.google.com/cloud-messaging/android/client

    replace the current service declaration that extends IntentService to the new GcmListenerService
    Refactor the current IntentService service implementation to use GcmListenerService

More details:

    A declaration of GcmListenerService, which enables various aspects of handling messages such as detecting different downstream message types, determining upstream send status, and automatically displaying simple notifications on the app’s behalf.
    A service that extends InstanceIDListenerService, to handle the creation, rotation, and updating of registration tokens.

tasomaniac:
This is the class where we receive the push messages. I am not 100% sure but it is not the place where the registration occurs.

Splaktar:
Right, it's not related to registration here. I'll take a look at that real quick.

Are we even using GCM for anything useful in the app? I'm seeing 10-20 registrations per day, but 0 messages per day.

tasomaniac:
We are supposed to send event notifications to the users but it is not
working for a looong time.

Splaktar:
We need to do something like this on registration:

    InstanceID instanceID = InstanceID.getInstance(FirstStartActivity.this);
                        String token = instanceID.getToken(BuildConfig.GCM_SENDER_ID,
                                GoogleCloudMessaging.INSTANCE_ID_SCOPE, null);

Then do the send to GdgXHub server in the service's listener. Details here: https://developers.google.com/cloud-messaging/android/client#sample-register

Examples here:
https://github.com/googlesamples/google-services/blob/master/android/gcm/app/src/main/java/gcm/play/android/samples/com/gcmquickstart/MyInstanceIDListenerService.java
https://github.com/googlesamples/google-services/blob/master/android/gcm/app/src/main/java/gcm/play/android/samples/com/gcmquickstart/RegistrationIntentService.java
https://github.com/googlesamples/google-services/blob/master/android/gcm/app/src/main/java/gcm/play/android/samples/com/gcmquickstart/MyGcmListenerService.java

Splaktar:
[...]

PR #644, opened 2016-06-07
Feature/442 Cloud Messaging

friedger:
This PR uses Google Service plugin 3.0 and the firebase configuration for FCM.

This is the preparation for more messaging logic. For now, we can't send messages to users per chapter, just the usual when in background messages. More depends on the backend.

screenshot_20160607-105354

### DIFF
--- a/app/google-services.json
+++ b/app/google-services.json
@@ -1,18 +1,16 @@
 {
   "project_info": {
-    "project_id": "api-project-429371117063",
     "project_number": "429371117063",
-    "name": "GDGX"
+    "firebase_url": "https://api-project-429371117063.firebaseio.com",
+    "project_id": "api-project-429371117063",
+    "storage_bucket": "api-project-429371117063.appspot.com"
   },
   "client": [
     {
       "client_info": {
         "mobilesdk_app_id": "1:429371117063:android:d42e7ba907a05c43",
-        "client_id": "android:org.gdg.frisbee.android",
-        "client_type": 1,
         "android_client_info": {
-          "package_name": "org.gdg.frisbee.android",
-          "certificate_hash": []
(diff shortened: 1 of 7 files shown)


friedger:
See #442

app/google-services.json
```diff
@@ -64,14 +62,7 @@
       ],
       "services": {
         "analytics_service": {
-          "status": 2,
-          "analytics_property": {
-            "tracking_id": "UA-42015512-1"
```
tasomaniac: I am afraid about these removal of analytics tracking ids. Is this json file generated by firebase?

app/src/main/res/values/tracker.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="ga_trackingId">UA-42015512-1</string>
```
tasomaniac: We had this implementation before. I don't know what is the best. I think if we use google's plugin, we should also use it for analytics. What do you think?

app/google-services.json
friedger: It is generated by firebase. This is the solution suggested here http://stackoverflow.com/a/37549781/243599 to use Firebase alongside GA.

app/src/main/res/values/tracker.xml
friedger: I didn't want to switch the analytics for this PR.

tasomaniac: I understand the reason now. Thank you for the stackoverflow link.

tasomaniac:
:+1:

tasomaniac:
:+1:

Issue #256, opened 2015-03-01
Greetings notification about new special event series. 


tasomaniac:
We now have times. It would be great to issue a notification when IO Extended starts.

tasomaniac:
Fixed by #645

PR #645, opened 2016-06-07
Feature/256 Greetings for tagged event series

friedger:
This PR generalizes the SummitNotificationHandler and allows to show alarm notifications for all users.

screenshot_20160607-225236

### DIFF
--- a/app/src/androidTest/java/org/gdg/frisbee/android/activity/NotificationTest.java
+++ b/app/src/androidTest/java/org/gdg/frisbee/android/activity/NotificationTest.java
@@ -1,0 +1,50 @@
+package org.gdg.frisbee.android.activity;
+
+import android.app.Notification;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.rule.ActivityTestRule;
+import android.support.test.runner.AndroidJUnit4;
+import android.support.test.uiautomator.By;
+import android.support.test.uiautomator.UiDevice;
+import android.support.test.uiautomator.UiObject2;
+import android.support.v4.app.NotificationManagerCompat;
+
+import org.gdg.frisbee.android.app.App;
+import org.gdg.frisbee.android.chapter.MainActivity;
+import org.gdg.frisbee.android.eventseries.NotificationHandler;
+import org.gdg.frisbee.android.eventseries.TaggedEventSeries;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+@RunWith(AndroidJUnit4.class)
+public class NotificationTest {
+    @Rule
+    public ActivityTestRule<MainActivity> rule = new ActivityTestRule<>(MainActivity.class);
+    private UiDevice mDevice;
+
+    @Before
+    public void setUp() throws Exception {
+        mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation());
+    }
+
+    @Test
+    public void testNotificationForEventSeries() {
+        int i = 0;
+        for (TaggedEventSeries series : App.getInstance().currentTaggedEventSeries()) {
+            Notification notification = new NotificationHandler(rule.getActivity(), series).createNotification();
+            NotificationManagerCompat.from(rule.getActivity()).notify(i, notification);
+
+            mDevice.openNotification();
+            mDevice.findObject(By.textContains(rule.getActivity().getString(series.getTitleResId())))
+                .click();
+            onView(withText(series.getDescriptionResId())).check(matches(isDisplayed()));
+        }
+    }
+}
--- a/app/src/main/java/org/gdg/frisbee/android/app/App.java
+++ b/app/src/main/java/org/gdg/frisbee/android/app/App.java
@@ -21,6 +21,7 @@ import android.location.LocationListener;
 import android.os.Bundle;
 import android.os.StrictMode;
 import android.support.annotation.NonNull;
+import android.support.v4.app.NotificationManagerCompat;
 
 import com.crashlytics.android.Crashlytics;
 import com.google.android.gms.analytics.GoogleAnalytics;
@@ -49,6 +50,7 @@ import org.gdg.frisbee.android.api.PlusApi;
 import org.gdg.frisbee.android.api.PlusApiFactory;
 import org.gdg.frisbee.android.api.PlusImageUrlConverter;
 import org.gdg.frisbee.android.cache.ModelCache;
+import org.gdg.frisbee.android.eventseries.NotificationHandler;
 import org.gdg.frisbee.android.eventseries.TaggedEventSeries;
 import org.gdg.frisbee.android.utils.CrashlyticsTree;
 import org.gdg.frisbee.android.utils.FileUtils;
@@ -58,6 +60,7 @@ import org.joda.time.DateTime;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Locale;
 
 import io.fabric.sdk.android.Fabric;
 import okhttp3.OkHttpClient;
@@ -194,12 +197,29 @@ public class App extends BaseApp implements LocationListener {
             Const.START_TIME_GCP_NEXT,
             Const.END_TIME_GCP_NEXT));
 
+        updateEventSeriesAlarms();
+
+    }
+
+    private void updateEventSeriesAlarms() {
+        if (isTranslationForAlarmsAvailable()) {
+            for (TaggedEventSeries eventSeries : currentTaggedEventSeries()) {
+                NotificationHandler notificationHandler = new NotificationHandler(this, eventSeries);
+                if (notificationHandler.shouldSetAlarm()) {
+                    notificationHandler.setAlarmForNotification();
+                }
+            }
+        }
+    }
+
+    private boolean isTranslationForAlarmsAvailable() {
+        return Locale.ENGLISH.getLanguage().equals(Locale.getDefault().getLanguage());
     }
 
     private void addTaggedEventSeriesIfDateFits(@NonNull TaggedEventSeries taggedEventSeries) {
         DateTime now = DateTime.now();
-        if (BuildConfig.DEBUG || (now.isAfter(taggedEventSeries.getStartDateInMillis())
-            && now.isBefore(taggedEventSeries.getEndDateInMillis()))) {
+        if (BuildConfig.DEBUG || (now.isAfter(taggedEventSeries.getStartDate())
+            && now.isBefore(taggedEventSeries.getEndDate()))) {
             mTaggedEventSeriesList.add(taggedEventSeries);
         }
     }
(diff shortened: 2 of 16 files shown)


app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
```diff
@@ -270,9 +272,11 @@ public void onConnected(Bundle bundle) {
     }
 
     private void initSummitNotificationHandler() {
-        NotificationHandler notificationHandler = new NotificationHandler(this);
-        if (notificationHandler.shouldSetAlarm()) {
-            notificationHandler.setAlarmForNotification();
+        for (TaggedEventSeries eventSeries : App.getInstance().currentTaggedEventSeries()) {
```
friedger: @tasomaniac this creates one alarm whenever is created a new activity, is that intended? Does it matter?

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeriesNotificationService.java
```diff
@@ -77,12 +77,14 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
...
         if (isOrganizer) {
+            TaggedEventSeries eventSeries = intent.getParcelableExtra(Const.EXTRA_TAGGED_EVENT);
+            NotificationHandler notificationHandler = new NotificationHandler(this, eventSeries);
             Notification notification = notificationHandler.createNotification();
 
             NotificationManager nm = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
             nm.notify(NOTIFICATION_ID, notification);
 
-            PrefUtils.setSummitNotificationSent(this);
+            PrefUtils.setTaggedEventSeriesNotificationSet(this, eventSeries);
```
friedger: Should the prefUtils handle the alarm instead of the notification?

app/src/main/java/org/gdg/frisbee/android/arrow/NotificationHandler.java
```diff
@@ -7,53 +7,51 @@
...
     }
 
     public void setAlarmForNotification() {
         AlarmManager am = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
 
         Intent intent = new Intent(SummitNotificationReceiver.ACTION_SUMMIT_NOTIFICATION);
+        intent.putExtra(Const.EXTRA_TAGGED_EVENT, eventSeries);
 
-        // Tuesday 8 AM PST in the morning
-        am.set(AlarmManager.RTC_WAKEUP, SUMMIT_DATE_TIME.getMillis(),
```
tasomaniac: We usually set the start date of an event series as the UTC midnight, right? Should we make this little bit more intelligent to set the morning in user's timezone?
Does it worth it?

app/src/main/java/org/gdg/frisbee/android/arrow/SummitNotificationReceiver.java
```diff
@@ -4,6 +4,9 @@
 import android.content.Intent;
 import android.support.v4.content.WakefulBroadcastReceiver;
 
+import org.gdg.frisbee.android.Const;
+import org.gdg.frisbee.android.eventseries.TaggedEventSeriesNotificationService;
+
 public class SummitNotificationReceiver extends WakefulBroadcastReceiver {
```
tasomaniac: Should we rename this as well then? 
And also, we can then move these classes to `eventseries` package.

app/src/main/java/org/gdg/frisbee/android/arrow/SummitNotificationReceiver.java
```diff
@@ -12,7 +15,8 @@ public SummitNotificationReceiver() {
 
     @Override
     public void onReceive(Context context, Intent intent) {
-        Intent serviceIntent = new Intent(context, SummitNotificationService.class);
+        Intent serviceIntent = new Intent(context, TaggedEventSeriesNotificationService.class);
+        serviceIntent.putExtra(Const.EXTRA_TAGGED_EVENT, intent.getParcelableExtra(Const.EXTRA_TAGGED_EVENT));
```
tasomaniac: The below line will handle passing all the extras. We don't need to pass this explicitly.

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
tasomaniac: `shouldSetAlarm` method below should prevent that from happening. It will set the alarm for only once, right? But this will be called in every Activity creation.
Should we then move this to the `App` class? That way we would do this only once per app start.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
```diff
@@ -157,6 +167,8 @@ public void writeToParcel(Parcel dest, int flags) {
         dest.writeInt(mLogoResId);
         dest.writeInt(mSpecialEventTheme);
         dest.writeInt(mDrawerId);
+        dest.writeInt(mGreetingsResId);
```
tasomaniac: You are missing these in the Parcelable constructor.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
```diff
@@ -32,6 +32,8 @@ public TaggedEventSeries createFromParcel(Parcel source) {
     private int mLogoResId;
     private int mSpecialEventTheme;
     private int mDrawerId;
+    private int mGreetingsResId;
+    private int mGreetingsTitleResId;
```
tasomaniac: Should these come from the theme as well?

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeriesNotificationService.java
tasomaniac: Definitely. I've implemented this quite rapidly. This is clearly a bug.

app/src/main/java/org/gdg/frisbee/android/arrow/NotificationHandler.java
friedger: If we don't make noise and just show a message, I think it is ok

app/src/main/java/org/gdg/frisbee/android/arrow/SummitNotificationReceiver.java
friedger: yes

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
friedger: Yes, or we define clever default values.

app/src/main/java/org/gdg/frisbee/android/utils/PrefUtils.java
```diff
@@ -182,11 +183,18 @@ public static void setFatalPlayServiceMessageShown(Context context) {
         prefs(context).edit().putBoolean(PREFS_FATAL_GOOGLE_PLAY_SERVICE, false).apply();
     }
 
-    public static boolean isSummitNotificationSent(Context context) {
-        return prefs(context).getBoolean(PREFS_SUMMIT_NOTIFICATION_SENT, false);
+    public static boolean isTaggedEventSeriesNotificationSet(Context context, TaggedEventSeries taggedEventSeries) {
+        long lastTimeSent = prefs(context).getLong(getKeyForEventSeries(taggedEventSeries), 0);
+        return lastTimeSent >= taggedEventSeries.getStartDateInMillis().getMillis();
```
tasomaniac: Next year when we increase the `startDateInMillis`, this will automatically return false. I like it. 👍

friedger:
@tasomaniac thanks for the review.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
tasomaniac: Oh, right. Maybe we don't need separate greeting for each event series. We can have a common one for all of them.

tasomaniac:
@friedger why did you close it? I think this is great!

app/src/main/java/org/gdg/frisbee/android/common/GdgActivity.java
friedger: If there was already an alarms with the same intent the previous is cancelled. So we are save.
See https://developer.android.com/reference/android/app/AlarmManager.html#set%28int,%20long,%20android.app.PendingIntent%29

Yes, to move it to `App`

tasomaniac: Yeah but creation of the alarm with intent and setting it can be costly, I think.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
friedger: good catch, fixed.

friedger:
The test does not pass yet due to missing IdleResourceHandler, see ci/integration-tests.

Fixes #256

friedger:
Issues address, can be reviewed and merged now.

app/src/main/java/org/gdg/frisbee/android/app/App.java
```diff
@@ -194,12 +196,29 @@ private void initTaggedEventSeries() {
...
+            for (TaggedEventSeries eventSeries : currentTaggedEventSeries()) {
+                NotificationHandler notificationHandler = new NotificationHandler(this, eventSeries);
+                if (notificationHandler.shouldSetAlarm()) {
+                    notificationHandler.setAlarmForNotification();
+                }
+            }
+        }
+    }
+
+    private boolean isTranslationForAlarmsAvailable() {
```
tasomaniac: Should we have this? I think English notification would be fine. Lots of users will miss the feature if we will have this.

app/src/main/java/org/gdg/frisbee/android/eventseries/TaggedEventSeries.java
```diff
@@ -134,11 +148,19 @@ public int getDrawerId() {
...
+        return mGreetingsTitleResId;
+    }
+
+    public DateTime getStartDate() {
         return mStartDateInMillis;
     }
 
-    public DateTime getEndDateInMillis() {
+    public DateTime getEndDate() {
         return mEndDateInMillis;
```
tasomaniac: Can you rename the field as well to `endDate`

tasomaniac:
@friedger to be quicker I just changed some naming. 

I just have the question regarding to the translation filtering.

app/src/androidTest/java/org/gdg/frisbee/android/activity/NotificationTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation());
+    }
+
+    @Test
+    public void testNotificationForEventSeries() {
+        int i = 0;
+        for (TaggedEventSeries series : App.getInstance().currentTaggedEventSeries()) {
```
tasomaniac: Wow! This is really cool. I didn't know this.

friedger:
@tasomaniac Feel free to to revert back the translation thing.  It was my
solution to the partial translated app problem.

tasomaniac:
I left it to you to get your opinion on this. What do you think? I am in between and confused :) 
But at the same time I think it would be better if all the users would benefit from this feature.

friedger:
Ideally, it would depend on the language of the chapter, e.g. in Brussels English is used.

For now, let's go with "English rules it all". In the future, we can use Remote Config for this and let the product manager decide on this.

friedger:
LGTM

tasomaniac:
Remote config makes sense a lot.

Issue #400, opened 2015-07-25
Events: Upcoming events should be first, past events after


Splaktar:
There should possibly be 2 sections to the events screen. One for future events and one for past events.

Currently I'm seeing the oldest event (from 2 months ago) first, which is the least interesting event to me. Then I have to scroll down below the fold to see my chapter's next event (which is the one I care the most about).

PareshMayani:
It's already in our issues list :) #311

Splaktar:
OK, I guess my search was just about the sorting so I didn't see that issue, but it should also fix this issue.

PareshMayani:
Yes that issue is about displaying 2 sections only, upcoming events section on top and past events section below to that!

PR #647, opened 2016-06-16
Endless Scrolling

lucasxas:
Add 'Endless Scrolling' to the GDG Events Fragment on the Main Activity.


tasomaniac:
Can you add screenshots or preferably gifs for the new UI? It would be great to see here.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -37,58 +41,33 @@ EventAdapter createEventAdapter() {
     }
 
     @Override
-    void fetchEvents() {
-        final DateTime now = DateTime.now();
-        DateTime start = now.minusMonths(2).dayOfMonth().withMinimumValue();
```
tasomaniac: Here we were giving the `start` date to the API which was about 2 months before the current date time. Where did this behavior go?

tasomaniac:
This is great but it has changed the sorting behavior. I don't know if the previous one was better but I think we should talk about it. 

In the previous implementation, the sorting was something like this: The future events were sorted in an ascending order. The nearest event is at the top. The past events at the bottom sorted in a descending order. The oldest event is at the bottom. 

That's actually why we wanted to separate them into 2 sections. 

Here are related issue #311 #400 #456

app/src/main/java/org/gdg/frisbee/android/api/GdgXHub.java
```diff
@@ -26,6 +26,13 @@
                                                @Path("start") final DateTime start,
                                                @Path("end") final DateTime end);
 
+    @GET("chapters/{chapterId}/events")
+    Call<PagedList<Event>> getChapterAllEventList(@Path("chapterId") final String chapterId,
```
friedger: If we have a use case for more parameters we should add them later

app/src/main/java/org/gdg/frisbee/android/common/GdgListFragment.java
```diff
@@ -321,13 +343,14 @@ private void ensureList() {
 
         if (!(rawList instanceof AdapterView)) {
```
friedger: Requires a check for `AbsListView` now. Or remove the check altogether.

friedger:
If they are in only list one have to be able to scroll into the past and into the future.

Splaktar:
Very interesting feature. Thank you very much for the contribution!

Looks like there is also some good discussion happening here in the review!

app/src/main/java/org/gdg/frisbee/android/api/GdgXHub.java
lucasxas: Yeah, can do that. Just thought it would be good to have a more generic method from start.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
lucasxas: We don't need this anymore right? The API endpoint will always give us the most recent events and as you scroll down you'll get more events.

lucasxas:
@tasomaniac I think if there's only one list having two different sorting methods makes it kinda of hard to the user to understand what's going on there. What do you think? 

If you asked me I'd say that even for two separated lists, we should be consistent with the sorting of the Events and keep them sorted in a descending order. I think that makes more sense when we talk about 'time' even if the event on top is one month and the second event is a week.

Or maybe we can automatically scroll the user to the nearest event so it would be on top in the first time you load the list and add something to show that there's more events if you scroll up. Kinda what Twitter does.

lucasxas:
Also, there's no New UI. The events will load in background as needed, depending on how many items are left in the list. For example if I'm at the top, and there's 20 items in the list, the Adapter won't load more events. But as soon as I scroll to the event 15 it will trigger a call to fetch the nextPage and add the result to the List.

lucasxas:
@tasomaniac @friedger are you guys waiting from anything from me here? :)

friedger:
:+1: Good for me @tasomaniac to merge.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
tasomaniac: Then, this is a behavior change. I am not sure about this. 
In the old behavior, the top event was always the next event. Now it is the latest event in the future.

tasomaniac:
@lucasxas I don't agree that we don't have a new UI here. This PR slightly changes the behavior and adds endless scrolling. We time to time reference to old PRs and issues. That's why we need a screenshot/gif of what you've done.

tasomaniac:
I really like the idea to scroll to the nearest event in the timeline. I think the next event should be the top visible in the list. Do you think that can be implemented as a part of this PR?

app/src/main/java/org/gdg/frisbee/android/Const.java
```diff
@@ -91,6 +91,8 @@
     public static final String CACHE_KEY_NEWS = "news2_";
     public static final String CACHE_KEY_PULSE = "pulse_";
 
+    //Order
```
tasomaniac: Can you remove this comment.

app/src/main/java/org/gdg/frisbee/android/common/GdgListFragment.java
```diff
@@ -116,6 +125,18 @@ public void onDestroyView() {
     public void onListItemClick(ListView l, View v, int position, long id) {
     }
 
+    /**
+     * This method will be called when new data needs to be
+     * appended to the List. Subclasses should override.
+     *
+     * @param page              The ListView where the click happened
```
tasomaniac: There is something wrong with param definitions. totalItemsCount cannot be the view clicked.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -37,58 +41,33 @@ EventAdapter createEventAdapter() {
...
-                    public void failure(Throwable error) {
-                        onError(R.string.fetch_events_failed);
-                    }
-
-                    @Override
-                    public void networkFailure(Throwable error) {
-                        onError(R.string.offline_alert);
-                    }
-                });
+            onListLoadMore(1, 0);
```
tasomaniac: `onListLoadMore(1, 0)` alone is not that understandable. Can you extract this into a separate method sayin something like `loadFirstPageOfEvents()`

app/src/main/java/org/gdg/frisbee/android/utils/EndlessScrollListener.java
```diff
@@ -0,0 +1,70 @@
...
+    // but first we check if we are waiting for the previous load to finish.
+    @Override
+    public void onScroll(AbsListView view, int firstVisibleItem, int visibleItemCount, int totalItemCount) {
+        // If the total item count is zero and the previous isn't, assume the
+        // list is invalidated and should be reset back to initial state
+        if (totalItemCount < mPreviousTotalItemCount) {
+            mCurrentPage = mStartingPageIndex;
+            mPreviousTotalItemCount = totalItemCount;
+            if (totalItemCount == 0) {
+                mLoading = true;
```
tasomaniac: Does this mean that initially we try to load the data using this callback? We are also loading it manually, right? How do we prevent loading twice?

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -143,4 +109,39 @@ private void splitEventsAndAddToAdapter(Collection<? extends SimpleEvent> events
...
+                    onError(R.string.fetch_events_failed);
+                }
+
+                @Override
+                public void networkFailure(Throwable error) {
+                    onError(R.string.offline_alert);
+                }
+            }
+        );
+        return true;
```
tasomaniac: We are always returning true here? How do we stop loading? When we are at the end of the data, we should stop loading more pages.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
```diff
@@ -143,4 +109,39 @@ private void splitEventsAndAddToAdapter(Collection<? extends SimpleEvent> events
         Collections.reverse(past);
         return new Pair<>(upcoming, past);
     }
+
+    @Override
+    protected boolean onListLoadMore(int page, int totalItemsCount) {
```
tasomaniac: the param `totalItemsCount` is not used in the method.

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
Splaktar: The top event needs to be the next event.

tasomaniac:
@lucasxas are you planning to continue on this? 

I have small comments + we should talk about the behavior change. The fastest way to merge this would be to pair on this. Would you like to pair and discuss in person?

lucasxas:
hey @tasomaniac yeah definitely, just being super busy, but I'm keen to finish this feature!

app/src/main/java/org/gdg/frisbee/android/Const.java
lucasxas: I think that comment helps keeping that Const class organized, thoughts?

app/src/main/java/org/gdg/frisbee/android/common/GdgListFragment.java
lucasxas: Fixed.

app/src/main/java/org/gdg/frisbee/android/utils/EndlessScrollListener.java
lucasxas: No, this just assumes that if the totalItems in the List is 0 we're loading the items somewhere else. It won't call the onLoadMore callback

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
lucasxas: Fixed!

lucasxas: Fixed!

app/src/main/java/org/gdg/frisbee/android/eventseries/GdgEventListFragment.java
lucasxas: Added a flag mReachedEndOfList, that is only true when the return data from the API returns an empty list OR if the list returned is smaller than the number of items we requested. So the next time we won't load again.

lucasxas: Added a flag mReachedEndOfList, that is only true when the return data from the API returns an empty list OR if the list returned is smaller than the number of items we requested. So the next time we won't load again.

Issue #648, opened 2016-06-16
Remove all games API related features


friedger:
This issue is about deciding how to proceed with the different authentication APIs, i.e. Google Sign In and Games.

We propose to remove all the Games API related features, i.e. Arrow and Achievements because there is no clean way to maintain both Games and Sign-In.

We saw people having problems to login because the Google Games ID setup and the Google Plus authorisation confuses users.

This issue was written together with @tasomaniac .

Splaktar:
I would raise this issue with the Google Games and Play Services teams. I would hate to lose the games features from the app. But I do understand that the APIs can be problematic.

tasomaniac:
We actually tried to reach out to them. These changes actually improves UX in actual games. It kinda removes the friction for games when you setup it once. 
The response that I could get from Googlers was kinda just confirms our situation. They said it's unfortunately how it is. 
There are also no real non game app out there that uses game apis

tasomaniac:
Fixed by #653

tasomaniac:
Google recently announced that the new sign in API finally works for games. 

https://android-developers.googleblog.com/2016/12/games-authentication-adopting-google.html

friedger:
Too late..

Am 18.12.2016 18:33 schrieb "Said Tahsin Dane" <notifications@github.com>:

tasomaniac:
😃

PR #653, opened 2016-09-10
Remove arrow, achievements and games

friedger:
This PR removes references to arrow, achievements and games. #648 describes in details why.


app/src/main/java/org/gdg/frisbee/android/app/GoogleApiClientFactory.java
```diff
@@ -26,12 +26,7 @@ private static GoogleApiClient createClient(Context context, boolean withSignIn)
             .addApi(AppIndex.API);
 
         if (withSignIn) {
-            Games.GamesOptions gamesOptions = Games.GamesOptions.builder()
-                .setRequireGooglePlus(true)
-                .setShowConnectingPopup(false).build();
-
             builder.addApi(Plus.API).addScope(Plus.SCOPE_PLUS_PROFILE)
-                .addApi(Games.API, gamesOptions)
                 .addApi(Drive.API).addScope(Drive.SCOPE_APPFOLDER);
```
tasomaniac: We are using this to save game state. I think we can remove it as well. If we want to store something we can use Firebase.

tasomaniac:
LGTM. I have only 1 comment. After fixing the lint errors it would be good to go. 

Do you think we should ask other's opinions on this? Nobody is using arrow game but I really liked achievements. Sad to see them go. :(

My personal opinion is :+1:

friedger:
We discussed that publicly, also in the last hangout. So, gone for now. I'd
say.

Let's create a arrow game app.

tasomaniac:
Cool. I couldn't attend the last hangout. 
Creating a separate game makes really sense. Let's do that.

tasomaniac:
LGTM

PR #654, opened 2016-09-12
sort gde categories in ascending order

PareshMayani:
I think GDE categories should be displayed in ascending order.

screenshot_20160912-144230


friedger:
:+1:

PR #660, opened 2016-09-14
Dependencies are updated

tasomaniac:
All the dependencies are updated except Play Services. They removed some deprecated classes regarding to Plus and GoogleSignIn. Once we move to new Auth, we can update it. 

The app tested manually.


tasomaniac:
@friedger I forgot to mention that in the desc. I didn't also do it because I thought we have a separate ticket for it. Let me append it quickly.

PR #661, opened 2016-09-14
Spinner redesign in MainActivity

### LINKED ISSUE #562: Redesign the spinner on the MainActivity
Redesign it to make it easier to find chapters in the huge list.

Replace the Spinner with a Button with a right carret that looks like a Spinner. 
On click of the button, re-use the implementation found in Settings.

tasomaniac:
Duplicated the behavior of `LocationListPreference` and created `ChapterSelectDialog`. 

Spinner and its adapter in `MainActivity` is removed. 
Instead we have now just a list of Chapters and a TextView. Clicking on it opens the Dialog with search. 

In xml, I just used `TextView` with `Spinner` style we had before. That turned out pretty good. We have the same look and feel. 

Right now, it still looks like a spinner. This can also be changed. I didn't want to spend too much time on design decisions. 

### Screenshots

Before | After
------ | ------
before| after

Fixes #562

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
+++ b/app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
@@ -90,7 +90,7 @@ public class MainActivity extends GdgNavDrawerActivity implements ChapterSelectD
     private boolean mFirstStart = false;
 
     private ChapterComparator locationComparator;
-    private TextView mSpinner;
+    private TextView chapterSwitcher;
     private String selectedChapterId;
     private ArrayList<Chapter> chapters;
 
@@ -115,7 +115,7 @@ public class MainActivity extends GdgNavDrawerActivity implements ChapterSelectD
         locationComparator = new ChapterComparator(PrefUtils.getHomeChapterIdNotNull(this),
             App.getInstance().getLastLocation());
 
-        initSpinner();
+        setupChapterSwitcher();
 
         if (savedInstanceState != null) {
             chapters = savedInstanceState.getParcelableArrayList(ARG_CHAPTERS);
@@ -368,12 +368,12 @@ public class MainActivity extends GdgNavDrawerActivity implements ChapterSelectD
         }
     }
 
-    private void initSpinner() {
+    private void setupChapterSwitcher() {
         Toolbar toolbar = getActionBarToolbar();
-        View spinnerContainer = LayoutInflater.from(this).inflate(R.layout.actionbar_chapter_selector, toolbar);
-        mSpinner = (TextView) spinnerContainer.findViewById(android.R.id.text1);
+        View container = LayoutInflater.from(this).inflate(R.layout.actionbar_chapter_selector, toolbar);
+        chapterSwitcher = (TextView) container.findViewById(android.R.id.text1);
 
-        mSpinner.setOnClickListener(new View.OnClickListener() {
+        chapterSwitcher.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
                 ChapterSelectDialog.newInstance(chapters, findChapterById(selectedChapterId))
@@ -386,7 +386,7 @@ public class MainActivity extends GdgNavDrawerActivity implements ChapterSelectD
         mViewPagerAdapter.setSelectedChapter(newChapterId);
         Chapter chapter = findChapterById(newChapterId);
         if (chapter != null) {
-            mSpinner.setText(chapter.toString());
+            chapterSwitcher.setText(chapter.toString());
         }
         if (!selectedChapterId.equals(newChapterId)) {
             Timber.d("Switching newChapterId!");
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -444,6 +428,11 @@ void launchAppInviteActivity(Intent intent) {
         startActivity(newIntent);
     }
 
+    @Override
+    public void onChapterSelected(Chapter selectedChapter) {
+        updateSelectionFor(selectedChapter.getGplusId());
```
friedger: here, you convert a chapter into an id and then back into a chapter. Should be made more performant.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
```diff
@@ -327,7 +323,7 @@ protected String getTrackedViewName() {
         } catch (IndexOutOfBoundsException e) {
             pageName = "";
         }
-        Chapter chapter = mChapterAdapter.findById(selectedChapterId);
+        Chapter chapter = findChapterById(selectedChapterId);
```
friedger: this could use chapter as a variable instead of searching for the chapter for each track event (not part of this PR though)

app/src/main/java/org/gdg/frisbee/android/widget/ChapterSelectDialog.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(Bundle savedInstanceState) {
+        View view = getActivity().getLayoutInflater()
+            .inflate(R.layout.view_location_list_preference, (ViewGroup) getView(), false);
+        ButterKnife.bind(this, view);
+
+        return new AlertDialog.Builder(getActivity())
+            .setNegativeButton(R.string.feedback_cancel, null)
```
friedger: feedback_cancel?

app/src/main/java/org/gdg/frisbee/android/widget/ChapterSelectDialog.java
```diff
@@ -0,0 +1,176 @@
...
+        args.putParcelable(EXTRA_SELECTED_CHAPTER, selectedChapter);
+        fragment.setArguments(args);
+        return fragment;
+    }
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        chapters = getArguments().getParcelableArrayList(EXTRA_CHAPTERS);
```
friedger: maybe it is better to get the chapters out of the cache instead of parceling in and out.

app/src/main/java/org/gdg/frisbee/android/chapter/MainActivity.java
tasomaniac: I think as a field we should also store `selectedChapter` as a whole object instead of just the id. I wanted to do it but I think it will blow up this PR.

app/src/main/java/org/gdg/frisbee/android/widget/ChapterSelectDialog.java
tasomaniac: I was just being lazy here. `feedback_cancel` is just cancel. That is something we add as a string for the feedback dialog. 
I think we should change this string to just `cancel`

app/src/main/java/org/gdg/frisbee/android/widget/ChapterSelectDialog.java
tasomaniac: Then, do I have to add the logic to get them from network if they are not available in the cache? Is that even be possible? 
If we are going to use it in Settings, it will definitely be possible.

tasomaniac:
@friedger Screen recordings are added

app/src/main/java/org/gdg/frisbee/android/widget/ChapterSelectDialog.java
tasomaniac: I just used `android.R.string.cancel` now 👍

friedger:
It is available in the cache as it is now used in the extra. For the edge
case where the cache is cleared due to low resources we can except the
extra delay.

Le 15 sept. 2016 00:28, "Said Tahsin Dane" <notifications@github.com> a
écrit :

tasomaniac:
I was just hesitant to add network calls into the dialog. It was because the dialog itself is so simple now. Let me look at it.

friedger:
Maybe not now.
But querying a repository should be easy ☺️

Le 15 sept. 2016 09:16, "Said Tahsin Dane" <notifications@github.com> a
écrit :

tasomaniac:
Most comments are addressed. 
No network call currently done in the dialog. It's just use the cache. It has a TODO in it as well. I think it will be fine. 
The other thing left is to adjust DialogPreference to use this Dialog. I want to do it in a separate PR. The problem is that it is not easy to use DialogFragment inside DialogPreference. I think the new preference support lib does it. I want to check that to see if it is easier with that.

friedger:
:+1:

PR #664, opened 2016-09-17
Move Constants to their related classes. 

tasomaniac:
My intention was to kill Const class. I believe constants should be per feature. They shouldn't be available publicly to the whole app. 

Now most of them are private. 

To do that TaggedEventSeries generation is moved to its own Factory. That also made Application class leaner.

### DIFF
--- a/app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
+++ b/app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
@@ -27,17 +27,17 @@ public class OrganizerChecker {
         initOrganizer();
     }
 
+    private void initOrganizer() {
+        mLastOrganizerCheck = mPreferences.getLong(PREF_ORGANIZER_CHECK_TIME, 0);
+        mCheckedId = mPreferences.getString(PREF_ORGANIZER_CHECK_ID, null);
+        mIsOrganizer = mPreferences.getBoolean(PREF_ORGANIZER_STATE, false);
+    }
+
     void resetOrganizer() {
         mPreferences.edit()
             .putBoolean(OrganizerChecker.PREF_ORGANIZER_STATE, false)
             .putLong(OrganizerChecker.PREF_ORGANIZER_CHECK_TIME, 0)
             .apply();
-    }
-
-    private void initOrganizer() {
-        mLastOrganizerCheck = mPreferences.getLong(PREF_ORGANIZER_CHECK_TIME, 0);
-        mCheckedId = mPreferences.getString(PREF_ORGANIZER_CHECK_ID, null);
-        mIsOrganizer = mPreferences.getBoolean(PREF_ORGANIZER_STATE, false);
     }
 
     public boolean isOrganizer() {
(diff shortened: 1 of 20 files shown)


tasomaniac:
I rebased this onto the latest develop. Should be fine now.

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
```diff
@@ -1,85 +1,90 @@
...
-    private boolean mIsOrganizer = false;
-    private long mLastOrganizerCheck = 0;
-    private String mCheckedId = null;
-    private SharedPreferences mPreferences;
 
-    public OrganizerChecker(SharedPreferences preferences) {
-        mPreferences = preferences;
-        initOrganizer();
-    }
+    private static final String PREF_ORGANIZER_CHECK_TIME = "pref_organizer_check_time";
```
friedger: pref keys resp. cache keys should be part of preferences resp. model cache to avoid clashes, they could also remain global IMHO

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
```diff
@@ -83,7 +85,7 @@ public void onActivityCreated(Bundle savedInstanceState) {
         if (Utils.isOnline(getActivity())) {
             setIsLoading(true);
             App.getInstance().getModelCache()
-                .getAsync(Const.CACHE_KEY_NEWS + plusId, new ModelCache.CacheListener() {
+                .getAsync(CACHE_KEY_NEWS + plusId, new ModelCache.CacheListener() {
```
friedger: As mentioned above cache keys should not be private

tasomaniac: Hmmm, I can revert that commit. I did cache keys part in 1 commit. 

Another way to prevent clash of keys is to put the name of the feature in front of the keys. What do you think?

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
tasomaniac: For pref keys, all of our keys are in PrefUtils. This is the only one that is separate. I think the reason is that this query/put multiple data at once.

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
friedger: I don't think we can enforce this properly. 

Shall we think of the keys as part of the cache or as part of the user of the cache? Difficult question. I tend towards the cache. So I'd prefer to go towards a cache that has knowledge of the model instead of cache users that have knowledge of all other cache users (at least of the convention that was agreed between them)

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
friedger: I don't understand that reason. Should go into PrefUtils then.

app/src/main/java/org/gdg/frisbee/android/chapter/NewsFragment.java
tasomaniac: That commit has reverted and then I moved all cache related consts to ModelCache class. Can you take a look?

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
tasomaniac: I'm thinking about this. Moving preferences related parts to PrefUtils is too much work right now. I don't know if it makes sense. 

One solution to this would be to use its own pref file, separate from the one used in `PrefUtils` That was there will not a any clash. 

Considering this was not in `PrefUtils` I think right now, it is much better.

PR #668, opened 2016-09-22
Firebase App Invite integration

### LINKED ISSUE #637: Migrate to Firebase Invites
https://firebase.google.com/docs/invites/android

tasomaniac:
First of all, this turned out bigger than I expected. Sorry for that. 🙏

This PR adds new Firebase Invites API. Firebase Invites uses Dynamic Links internally which is where the magic occurs.

App Invite, on top of that, a UI where you can choose your friends to share. But it only includes sharing via SMS and Email. I really didn't like it. I think it is too restrictive. 

Instead, I create dynamic links directly. Which then, user can choose whatever app they want to share. 

I'm using https://gdg.events/invite/?sender=gplusid as link structure. This link will be opened if it is clicked in a computer. 

If the app is already installed, it just opens the app. 

If not, it installs the app and then in the login page, I put the picture and a welcome message there. 
Since we have more info there now, I quickly add landscape version of the page. 

## Screenshots

| Portrait | Landscape |
| ------- | ----------- |
| screenshot_1474578097 | screenshot_1474580562 |

Fixes #637


app/build.gradle
```diff
@@ -136,10 +136,11 @@ dependencies {
     compile "com.android.support.test.espresso:espresso-idling-resource:$rootProject.espressoVersion"
 
     //Play Services
+    compile "com.google.android.gms:play-services-auth:$rootProject.playServicesVersion"
```
friedger: Why do you need auth? It is not mentioned in the firebase integration documentation?

app/src/main/java/org/gdg/frisbee/android/about/AboutActivity.java
```diff
@@ -72,17 +75,26 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
             return true;
         }
 
         return super.onOptionsItemSelected(item);
     }
 
+    private void shareAppInviteLink() {
+        AppInviteLinkGenerator linkGenerator = AppInviteLinkGenerator.create();
+        String gplusId = PlusUtils.getCurrentPersonId(getGoogleApiClient());
+        HttpUrl appInviteLink = gplusId != null
```
friedger: AboutActivity should not decide on how to handle not signed users. This should go into the linkGenerator.

app/src/main/java/org/gdg/frisbee/android/common/GdgNavDrawerActivity.java
```diff
@@ -303,6 +312,19 @@ private void navigateTo(Class<? extends GdgActivity> activityClass, Bundle addit
         mDrawerLayout.closeDrawers();
     }
 
+    private void shareAppInviteLink() {
```
friedger: Duplicate code, same as in AboutActivity

app/src/main/java/org/gdg/frisbee/android/onboarding/FirstStartStep2Fragment.java
```diff
@@ -18,25 +18,65 @@
...
+            if (inviteSender != null) {
+                loadSender(inviteSender);
+            }
+        }
+    }
+
+    @Override
+    public void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+        outState.putString(INVITE_SENDER,  inviteSender);
```
friedger: :+1:

app/src/main/java/org/gdg/frisbee/android/onboarding/FirstStartStep2Fragment.java
```diff
@@ -62,6 +102,39 @@ public void onSkippedSignIn() {
...
+        this.inviteSender = inviteSender;
+        if (!isContextValid()) {
+            return;
+        }
+
+        App.getInstance().getPlusApi()
+            .getPerson(inviteSender)
+            .enqueue(new Callback<Person>() {
+                @Override
+                public void success(Person sender) {
```
friedger: If we fail to load the sender then the invitation process stops here. We should still show the inviteContainer with a dummy image and a dummy name ("a friend" or so)

app/src/main/res/values/strings.xml
```diff
@@ -164,15 +164,12 @@
...
-    <string name="action_deep_link" translatable="false">
-        org.gdg.frisbee.android.DEEP_LINK_ACTION
-    </string>
+  <string name="invitation_message">Download GDG Android App and discover Google Developers Group content while being on the go. %s</string>
   <string name="invitation_error_message">App Invite was unsuccessful.</string>
   <string name="thanks_for_feedback">Thank you for your Feedback!</string>
   <string name="retry">Retry</string>
   <string name="title_event_series_greetings">New events coming up!</string>
   <string name="event_series_greetings">Learn more about %1$s and see events around the world!</string>
+  <string name="cd_invitation_sender_profile_image">Invitation sender profile image</string>
```
friedger: Did you upload the new strings to crowdin?

app/build.gradle
tasomaniac: The piece of code that we have in Settings it's not working anymore. The class was removed. I first deleted that and realized that I can just add this dependency.  The reason was the upgrade of the gms version.

app/src/main/java/org/gdg/frisbee/android/onboarding/FirstStartStep2Fragment.java
tasomaniac: Totally makes sense. We should do it.

app/src/main/res/values/strings.xml
tasomaniac: Strings are already upload/updated in crowdin.

tasomaniac:
@friedger comments are addressed. 👍

app/src/main/java/org/gdg/frisbee/android/onboarding/Invite.java
```diff
@@ -0,0 +1,38 @@
+package org.gdg.frisbee.android.onboarding;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+
+class Invite implements Parcelable {
```
friedger: Why do you add this class? Is this "just" for giving String a meaning? Then `sender` should be `gplusIdOfSender` or so

tasomaniac:
I did that to both provide a meaningful name and to handle 3 cases in 1
object.
String is not enough. We have 3 states.
No invite, invite without sender, invite with sender.

PR #682, opened 2016-10-04
refactor: Speed up build times

esoxjem:
Enable buildCache for incremental builds


tasomaniac:
Hey this is an experimental flag and can be enabled locally by setting this flag on global Gradle properties. 
That's why I will close it. You can document this in a wiki under a possible title like speeding up your builds.

esoxjem:
@tasomaniac I think only collaborators can edit wikis?

PR #684, opened 2016-10-04
refactor: update signingConfigs

esoxjem:
set settings/debug.keystore as default for debug builds for G+ sign-in to work


friedger:
The debug config is already defined in frisbee.gradle

esoxjem:
Oh okay. Then I guess this line can be removed from the wiki: "Furthermore, you need to use the settings/debug.keystore for signing as it is whitelisted for Google Plus Sign-in."

PR #690, opened 2016-10-07
616: Use new Google Auth API for signIn

tasomaniac:
This PR uses the new Google Auth API. 

- With this we removed couple deprecated methods/classes from Plus API. 
- GoogleApiClientFactory with special ApiClient just for login.
- Removed all login related stuff from `GdgActivity` Login should not be done in the base class. It is now explicitly done by using the special ApiClient. 
- Removed custom G+ login assets. Instead we are using the button from Google. Unfortunately recently, they removed all G+ branding. Now it is not possible to have a G+ button. It is only Google button.
- SharedPrefs is set only when the login is successful. In the past, we were setting it true when sign-in was clicked. Which means that the user wants login. And then it was handled automatically in GdgActivity. But now, we are handling it explicitly.

screenshot_1475878299


app/src/main/java/org/gdg/frisbee/android/app/GoogleApiClientFactory.java
```diff
@@ -1,42 +1,62 @@
...
-            .addApi(AppIndex.API);
+    public static GoogleApiClient createForSignIn(Context context) {
+        return createBuilderForSignIn(context)
+            .build();
+    }
 
-        if (withSignIn) {
-            builder.addApi(Plus.API).addScope(Plus.SCOPE_PLUS_PROFILE);
-        }
+    public static GoogleApiClient createForSignIn(FragmentActivity fragmentActivity,
```
esoxjem: Shouldn't the first param be `Context`?

app/src/main/java/org/gdg/frisbee/android/utils/PlusUtils.java
```diff
@@ -1,28 +1,36 @@
...
     @Nullable
-    public static String getCurrentPersonId(GoogleApiClient apiClient) {
-        Person plusPerson = null;
-        if (apiClient.isConnected() && apiClient.hasConnectedApi(Plus.API)) {
-            plusPerson = Plus.PeopleApi.getCurrentPerson(apiClient);
+    public static GoogleSignInAccount getCurrentAccount(Context context) {
+        GoogleApiClient client = GoogleApiClientFactory.createForSignIn(context);
+
+        OptionalPendingResult<GoogleSignInResult> result = Auth.GoogleSignInApi.silentSignIn(client);
+        if (!result.isDone()) {
```
tasomaniac: When isDone is true, it means that everything is ready and locally saved. So it should be instant. 

And the beat thing is, it will be false automatically when user goes into setting in web and revokes permissions for our app.

app/src/main/java/org/gdg/frisbee/android/app/GoogleApiClientFactory.java
tasomaniac: enableAutoManage method requires FragmentActivity unfortunately

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
```diff
@@ -45,40 +43,44 @@ public boolean isOrganizer() {
...
+
+            @Override
+            public void failure(Throwable error) {
+                isOrganizer = false;
+                responseHandler.onErrorResponse();
+            }
+        });
+    }
+
+    private boolean isLastOrganizerCheckValid(String currentId) {
```
lucasxas: This method name is a bit strange.. just looking at the code and the name I'm not sure what we're doing here.

app/src/main/java/org/gdg/frisbee/android/utils/PlusUtils.java
friedger: What happens when it is false? When do you re-query?

tasomaniac: I think it will just return null because we do not have permission anymore. 
In that case we need set the signIn pref to false so that it will not be checked anymore in the settings screen. I can do that. 

Do you think should we just log out or should we re-try to sign in? 

In the old app, I think it was annoying that it was retrying so consistently.

app/src/main/java/org/gdg/frisbee/android/app/OrganizerChecker.java
tasomaniac: It was harder to understand before. I think @friedger 's name suggestion makes sense.

app/src/main/java/org/gdg/frisbee/android/utils/PlusUtils.java
```diff
@@ -1,28 +1,36 @@
...
+
+import org.gdg.frisbee.android.app.GoogleApiClientFactory;
 
 public class PlusUtils {
 
-    private PlusUtils() {
+    @Nullable
+    public static String getCurrentPersonId(Context context) {
+        GoogleSignInAccount account = getCurrentAccount(context);
+        return account != null ? account.getId() : null;
```
tasomaniac: Yep, I've checked. It is the G+ id.

tasomaniac:
I went ahead and merged this. Because future stuff will be depending on this. 

I've tested it pretty heavily, it looks like it works very well. I will now push this to internal alpha build.

PR #712, opened 2016-12-24
Update build tools and dependencies

tasomaniac:
Update build tools and dependencies


tasomaniac:
Play services was not updated because Google's `app-indexing` was removed in the latest artifacts. We need to migrate to Firebase App Indexing.

PR #713, opened 2016-12-25
Integrate firebase test lab

tasomaniac:
I've spent some time to integrate Firebase Test Lab. 
They now have some free quota every day. I think that would be enough for us. 

I've wrote a small script that only runs to upload test APKs to Firebase on some certain cases. That's now after every push on develop.

That would be more than enough for us. 

Also I realized that our tests just fail because of a Dialog covering the screen on Christmas season. :) Although that dialog is awesome, I disabled it for tests by overriding its value.


tasomaniac:
Let's see if it will run after merge.

PR #718, opened 2017-04-14
Remove outdate info

ligi:
This is no more needed since AS 2.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,6 @@ Read the [Development Guide](https://github.com/gdg-x/frisbee/wiki/Developer-Doc
 
 When sending pull requests please make sure to enable EditorConfig in Android Studio -> Settings -> Editor -> Code & Style -> EditorConfig.
 
-#### Speeding up debug builds
-
-The project uses multidex. To speed up the builds you need to set `minSdkLevel` to 21 and above. Our project uses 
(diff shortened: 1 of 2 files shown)


tasomaniac:
AFAIK, this optimization is only available in AS 2.3 when instant run is enabled right? Do you have more information around this? 

Also if we want to remove this, we should remove the code that checks and modifies the min sdk version.

ligi:
It was mentioned in the androidstudygroup slack. I removed this workaround from my apps and don't see a performance impact - but I am also using instant run.
Also changed the build.gradle here now - though I do not think both changes really need to go together. And currently the markdown in the README is broken - deleting this part also fixes this ..-)

tasomaniac:
Thank you for your contribution and deleting the unnecessary part from `build.gradle` too. 👍

tasomaniac:
Merging this was not a big problem but I just want to update. 
According to [ADB Podcast Gradleweiss](http://androidbackstage.blogspot.de/2017/04/episode-64-gradleweiss.html) it is added in build tools 2.4 and only for instant run. It is not stable currently but will be soon.

ligi:
Really strange - was just searching for the conversation to check if I understood something wrong there - but I would still interpret it the same way - and this guy is on the AS team:

selection_039

Have not yet listened to this podcast - but I am a bit confused now.

